PR #51, opened 2017-08-02
feat(loader): The loader component. Full size and inline.

keepitterron:
loader


src/index.scss
```diff
@@ -58,6 +58,7 @@ $ns: 'sb-' !default;
 @import 'components/preview';
 @import 'components/icons';
 @import 'components/select';
+@import 'components/loader';
 
-// Trumps
+// Trumps @>:^0
```
dine: 🤡

src/components/_loader.scss
```diff
@@ -0,0 +1,111 @@
+// Loader
+//
+// Markup:
+// <div class="sb-loader sb-loader--inline">
```
dine: If we can make the markup the same and the only change is the modifier class, we could add `.sb-loader--inline`declared after the markup, like we do in `buttons.scss` for example.

src/components/_loader.scss
```diff
@@ -0,0 +1,111 @@
...
+$sb-loader-size: sb-px2rems(200px);
+$sb-loader-padding-inline: sb-px2rems(13px);
+$sb-loader-animation-duration: 0.8s;
+
+.#{$ns}loader {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+
+  &-label {
```
dine: We are using [BEM](http://getbem.com/naming/), so the elements should have a `__` instead of `-`.

src/components/_loader.scss
```diff
@@ -0,0 +1,111 @@
...
+  &--inline {
+    flex-direction: initial;
+    background: $sb-color-white;
+    border: 1px solid $sb-color-light-3;
+    width: $sb-loader-size;
+    border-radius: 2px;
+    padding: 0 $sb-loader-padding-inline;
+
+    .#{$ns}loader-spinner {
+      margin-left: auto;
```
dine: If we use the approach of same markup with modifier class, here we will have to declare here the new size / position for the spinner.

src/components/_loader.scss
```diff
@@ -0,0 +1,114 @@
...
+  &__spinner--secondary {
+    fill: $sb-color-light-3;
+  }
+
+  &--inline {
+    flex-direction: initial;
+    background: $sb-color-white;
+    border: 1px solid $sb-color-light-3;
+    width: $sb-loader-size;
+    border-radius: 2px;
```
endritbajo: Here we can use `$sb-radius`

src/components/_loader.scss
```diff
@@ -0,0 +1,114 @@
...
+//
+// Styleguide 16.
+
+$sb-loader-size: sb-px2rems(200px);
+$sb-loader-spinner-size: sb-px2rems(16px);
+$sb-loader-padding-inline: sb-px2rems(13px);
+$sb-loader-animation-duration: 0.8s;
+
+.#{$ns}loader {
+  display: flex;
```
endritbajo: What about browser support, does it work on IE < 11?

src/components/_loader.scss
```diff
@@ -0,0 +1,114 @@
+// Loader
+//
+// Markup:
+// <div class="sb-loader sb-loader--inline">
+//   <span class="sb-loader__label">Loading</span>
+//   <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="51" height="50" viewBox="0 0 51 50" class="sb-loader__spinner">
```
beerecca: Lots we can do to tidy up this svg. Take a look at https://jakearchibald.github.io/svgomg/ to start with (once I guess we get a new version from Gal).

src/components/_loader.scss
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  &__spinner {
+    animation-duration: $sb-loader-animation-duration;
+    animation-name: sb-rotate-loader;
+    animation-iteration-count: infinite;
+    animation-timing-function: linear;
+  }
+
+  &__spinner--primary {
```
beerecca: This isn't a modifier so shouldn't have `--`. Can instead create a new base - `sb-spinner` (not nested inside the loader class), with `sb-spinner__primary` and `sb-spinner__secondary`.

src/components/_loader.scss
beerecca: Check accessibility on this, especially with a screen reader - should at least have aria-hidden on the loader I think?

src/components/_loader.scss
```diff
@@ -0,0 +1,114 @@
...
+$sb-loader-padding-inline: sb-px2rems(13px);
+$sb-loader-animation-duration: 0.8s;
+
+.#{$ns}loader {
+  display: flex;
+  flex-direction: column-reverse;
+  align-items: center;
+
+  &__label {
+    padding: $sb-loader-padding-inline 0;
```
beerecca: Can we not use an existing padding class? Same below

src/components/_loader.scss
```diff
@@ -0,0 +1,114 @@
...
+//           </g>
+//         </g>
+//       </g>
+//     </g>
+//   </svg>
+// </div>
+//
+// Styleguide 16.
+
+$sb-loader-size: sb-px2rems(200px);
```
beerecca: Same issue we had with the dropdown - I don't think we should fix this. I think it should be 100%, then the user can put it in whatever container (or grid column) they want. (If we do stick with this though it should be `sb-inline-loader-width`)

src/components/_loader.scss
beerecca: Should remove the fixed width and height from the svg too.

keepitterron:
Did a major refactoring after getting more familiar with the tool and processes and reading through your comments.
I also had to recreate from scratch the svg spinner.
Also rewrote git history by rebasing and removing obsolete commits, sorry about that :P

src/components/_loader.scss
keepitterron: I'll do a style check with Gal tomorrow. I also think this should be just 100% width.

src/components/_loader.scss
keepitterron: I couldn't find a `13px` spacing generic mixin/class used elsewhere?

keepitterron:
I should have addressed all your comments, if not pls bring it up.
- old browser support
- aria-hidden
- separate `sb-spinner` class
- proper use of BEM
- same markup
- smooth, optimized svg

src/index.scss
```diff
@@ -58,6 +58,7 @@ $ns: 'sb-' !default;
 @import 'components/preview';
 @import 'components/icons';
 @import 'components/select';
+@import 'components/loader';
 
 // Trumps
```
dine: no more  @>:^0? ☹️

keepitterron: oh nooooo, i lost it with `git rebase` :(

src/components/_loader.scss
```diff
@@ -0,0 +1,89 @@
+// Loader
+//
+// Markup:
+// <div class="sb-loader {{modifier_class}}">
+//   <svg viewBox="0 0 50 50" class="sb-spinner" aria-hidden="false">
```
beerecca: Should be aria hidden true. Because it won't mean anything to a blind person. Only the loading label will. Maybe do a bit of research on this, what's the best approach for a spinner?

beerecca: Still lots that can be cleaned up with the svg. Shouldn't need all those ids, groups, stroke rules, widths and heights etc.

src/components/_loader.scss
```diff
@@ -0,0 +1,89 @@
...
+//       </g>
+//   </svg>
+//   <span class="sb-loader__label">Loading</span>
+// </div>
+//
+// .sb-loader--inline - Inline loader
+//
+// Styleguide 16.
+
+$sb-loader-size--inline: sb-px2rems(200px);
```
beerecca: Want to change this before we merge it

beerecca: Want to change this before we merge it

src/components/_loader.scss
```diff
@@ -0,0 +1,89 @@
...
+$sb-loader-spinner-size: sb-px2rems(50px);
+$sb-loader-spinner-size--inline: sb-px2rems(16px);
+$sb-loader-padding: sb-px2rems(13px);
```
beerecca: Couldn't comment on your comment for some reason  - but I really don't like the idea of a random new padding. Will one of the others still look good?

src/components/_loader.scss
keepitterron: already on it :)

src/components/_loader.scss
keepitterron: yes, it's on my list of things to discuss with gal. our padding small is `11px`, don't see much difference there.

keepitterron:
Last update.
- SVG only has properties it really needs
- did a spacing check with Gal, using 11px/16px now
- parent will decide the inline loader size (100%)

PR #54, opened 2017-08-08
Show colors in styleguide

keepitterron:
Reviewers: @beerecca @kdabo @endritbajo 

colors

There are two things that annoyed me while working on this project that I'd like to fix:
- having to remember where we're at with styleguide refs: `// Styleguide 1.2.3`
- not being able to see the colors in the styleguide and having to look for them

I did a bit of hacking to bring the colors into the styleguide and I replaced the numbers with labels when using references `// Styleguide 1.2.3` => `// Styleguide colors.charts.blue`
It's easier this way to deal with it.

What do you guys think?


docs-builder/kss-assets/kss.js
```diff
@@ -51,3 +51,27 @@
   new KssStateGenerator;
 
 }).call(this);
+
+
+// colors.
+(function(){
+  var parameters = document.querySelectorAll('.kss-parameters');
```
kdabo: should we use `const` instead of `var`?

src/components/_buttons-group.scss
```diff
@@ -10,7 +10,7 @@
 //   <button class="sb-btn" disabled>Disabled</button>
 // </div>
 //
-// Styleguide 10.
+// Styleguide buttons-group
```
dine: Didn't know we could use strings instead of numbers, much better! 👍

docs-builder/kss-assets/kss.scss
```diff
@@ -764,3 +764,42 @@ $kss-vertical-rhythm      : $kss-font-size * 1.5;
     }
   }
 }
+
+// Custom styles
+.kss-colors-container {
+  margin: 0;
+  padding-left: 0;
+  display: flex !important;
```
kdabo: why do we need important here?

keepitterron: 'cause i'm hacking the kss generated style I'm not in control of.

docs-builder/kss-assets/kss.js
keepitterron: this is a third party file and it's old, it has `vars` and such all over the place.
it doesn't really matter.

PR #55, opened 2017-08-08
feat(preview): add preview frame and label

beerecca:
@usabilla/mallard


src/components/_preview.scss
```diff
@@ -16,8 +16,29 @@
...
+  &__label {
+    position: absolute;
+    left: 50%;
+    top: 50%;
+    transform: translate(-50%, -50%);
+  }
+
+  &__frame {
+    position: absolute;
+    top: sb-px2rems(20px);
```
keepitterron: why `20px`? isn't our spacing [...16,23...]?

beerecca: hmm good point... i got this from the grid margin. So maybe that needs to be 23 too..?

keepitterron: Ah yes,
we should revisit the grid system now that the styleguide is updated and we have one example of usage.
Should we open a ticket for it?

beerecca: There's one on the styleguide jira project STYLE-40, I just added some notes to it.

PR #56, opened 2017-08-08
feat(deploy): deploy stylabilla to s3

beerecca:
@endritbajo @kdabo 

Waiting on gijs to add s3 credentials to circle before merging.


beerecca:
Which is now done!

PR #58, opened 2017-08-08
Flam 629/dropdown component

kdabo:
Reviewers: @endritbajo @beerecca 

Added basic dropdown element, together with disabled style as well as invalid error. 

screen shot 2017-08-08 at 17 11 08
screen shot 2017-08-08 at 17 11 17
screen shot 2017-08-08 at 17 11 25


src/components/_dropdown.scss
```diff
@@ -0,0 +1,124 @@
...
+    text-align: left;
+    z-index: 1;
+
+    &:hover,
+    &:focus {
+      box-shadow: none;
+    }
+  }
+
+  &-menu {
```
endritbajo: The menu needs to be position absolute as it needs to be floating over other elements

src/components/_dropdown.scss
```diff
@@ -0,0 +1,124 @@
...
+    display: none;
+    font-family: $sb-base-font-family;
+    font-weight: 400;
+    font-size: sb-px2rems(13px);
+    border: 1px solid $sb-color-light-3;
+    border-radius: 0 0 sb-px2rems($sb-radius) sb-px2rems($sb-radius);
+    background-color: $sb-color-white;
```
endritbajo: Maybe this is redundant?

src/components/_dropdown.scss
```diff
@@ -0,0 +1,124 @@
...
+//
+// .sb-dropdown--open - open dropdown
+//
+// Markup:
+//  <div class="sb-form-control">
+//    <label class="sb-label sb-padding-bottom-xs">Choose an option</label>
+//    <div class="{{modifier_class}}">
+//      <button class="sb-dropdown__button sb-input">
+//        Choice 1
+//        <span class="sb-select__arrow">
```
endritbajo: Maybe we can re-use the `sb-input` classes:

```html
   <div class="sb-input-wrapper">
     <button class="sb-input">Choice 1</button>
     <svg class="sb-icon" ...>
        ....
     </svg>
   </div>
```

src/components/_dropdown.scss
kdabo: It is not. Without it, there is no border on the list item and the font that appears is Milo. 
Also, with no `background-color` defined, the list is transparent.

src/components/_dropdown.scss
kdabo: `sb-icon` class is this:
`  .sb-icon {
    height: 1rem;
    width: 1rem;
} `, whereas I am using `sb-select__arrow` class which is used by `select` component and is following: 
`position: absolute;
  right: .5625rem;
  top: 50%;
  transform: translateY(-50%);
`

endritbajo: We should try to use the same icon system. We want to be able to use it on the `Icon` component in `react-stylabilla`. We should change it in select component as well

src/components/_dropdown.scss
dine: According to [BEM](http://getbem.com/naming/), this would be an element inside the block `.sb-dropdown`, right? So instead of `&-menu` we would have `&__menu` here.

src/components/_dropdown.scss
```diff
@@ -0,0 +1,126 @@
...
+      box-shadow: none;
+    }
+  }
+
+  &-menu {
+    display: none;
+    position: absolute;
+    width: 100%;
+    font-family: $sb-base-font-family;
+    font-weight: 400;
```
dine: The default font-weight is 400, so maybe we don't need to declare it here?

src/components/_input.scss
```diff
@@ -79,12 +79,8 @@
     padding-right: sb-px2rems(23px);
   }
 
-  .#{$ns}icon {
+  .#{$ns}icon--input {
```
dine: We have this already implemented in `react-stylabilla`. We need to be careful with this [breaking changes](https://github.com/usabilla/stylabilla/blob/master/CONTRIBUTING.md). And we need to update the commit message accordingly.

dine: Sorry, now that I understand we are using the same styles for validation icons and the dropdown arrow. Even being at the same place right now, they have different meanings/purposes. We could want in the future to replace the position of the validation icons without affecting the arrow for example.

src/components/_select.scss
```diff
@@ -11,7 +11,7 @@
 //        <option>Choice 2</option>
 //        <option>Choice 3</option>
 //      </select>
-//      <span class="sb-select__arrow">
+//      <span class="sb-icon sb-icon--input">
```
dine: The select and dropdown arrow could be the same. And the styles related to the icon only can be declared in a modifier, as in `--arrow` (things like width, height, fill). But the styles related to the behaviour (like position, right, etc) I believe are more like an element, so `__arrow`? Also, we need to also add the svg to `_icons.scss`

src/components/_dropdown.scss
```diff
@@ -0,0 +1,126 @@
...
+//  <div class="sb-form-control">
+//    <label class="sb-label sb-padding-bottom-xs">Choose an option</label>
+//    <div class="{{modifier_class}}">
+//      <button class="sb-dropdown__button sb-input">
+//        Choice 1
+//        <span class="sb-icon sb-icon--input">
+//          <svg width="8" height="4" viewBox="0 0 8 4"><path d="M0 0l3.5 4L7 0H0"></svg>
+//        </span>
+//      </button>
+//      <ul class="sb-dropdown-menu" role="menu">
```
katia-agna: here the roles are not really really representative. We don't really understand with the screen reader. 
if you change the role of `<ul>` with `role="listbox"` and each `<li>` can have a `role="option"` like the the screen reader is able to say "choice 1 (1 to 3)"

src/components/_dropdown.scss
```diff
@@ -0,0 +1,126 @@
...
+//        <li class="sb-dropdown-menu__item" tabindex="0" role="menuitem">Choice 1</li>
+//        <li class="sb-dropdown-menu__item" tabindex="0" role="menuitem">Choice 2</li>
+//        <li class="sb-dropdown-menu__item" tabindex="0" role="menuitem">Choice 3</li>
+//      </ul>
+//    </div>
+//    <p class="sb-form-control__validation">Error message here.</p>
```
katia-agna: We should add an aria-invalid linked to this `<p`> when error. But we may do that with Javascript part right?

src/components/_dropdown.scss
kdabo: that is true @dine ! I noticed a lot of inconsistency here as it is not everywhere used like that.

beerecca: Argh it should be - can you show me where BEM is used incorrectly?

src/components/_dropdown.scss
```diff
@@ -0,0 +1,134 @@
...
+      background-color: $sb-color-blue-4;
+    }
+  }
+
+  &--open &__menu {
+    display: block;
+  }
+}
+
+.#{$ns}input-wrapper .#{$ns}icon {
```
endritbajo: This is not scoped anymore so being declared globally it will affect all `sb-icon` on the project.

src/components/_dropdown.scss
```diff
@@ -0,0 +1,134 @@
...
+  display: inline;
+  z-index: 2;
+
+  &--disabled {
```
endritbajo: We don't need to add another property, we could just listen to
```sass
.sb-input:disabled + .sb-icon {
  fill: $sb-color-light-3;
}
```

src/components/_input.scss
```diff
@@ -23,10 +23,6 @@
   font-size: sb-px2rems(13px);
   color: $sb-color-dark-3;
 
-  &::placeholder {
```
endritbajo: Why did we remove this?

src/components/_input.scss
```diff
@@ -10,7 +10,7 @@
 //   </div>
 // </div>
 //
-// Styleguide input
+// Styleguide 7.
```
endritbajo: Is there a reason we changed it to a number?

src/components/_dropdown.scss
kdabo: that was my intention as I couldn't reuse it. it is for input-icons.

beerecca: It looks like the styles you want to apply here are specific to the dropdown. So I would put the svg inside the button, refer to it with .sb-dropdown__button .sb-icon and remove sb-input-wrapper altogether

src/components/_dropdown.scss
```diff
@@ -0,0 +1,134 @@
...
+//        </button>
+//          <svg class="sb-icon" width="16" height="8" viewBox="0 0 16 8"><path d="M0 0l3.5 4L7 0H0"></svg>
+//      </div>
+//      <ul class="sb-dropdown__menu" role="listbox">
+//        <li class="sb-dropdown-menu-item" tabindex="0" role="option">Choice 1</li>
+//        <li class="sb-dropdown-menu-item" tabindex="0" role="option">Choice 2</li>
+//        <li class="sb-dropdown-menu-item" tabindex="0" role="option">Choice 3</li>
+//      </ul>
+//    </div>
+//    <p role="alert" class="sb-form-control__validation">Error message here.</p>
```
beerecca: If this needs a role of alert, then i guess they all should? Or is only when it appears, in which case we should add documentation for the js developer. Either way, can we make it consistent across the other error messages

src/components/_select.scss
```diff
@@ -5,34 +5,22 @@
 // Markup:
 //  <div class="sb-form-control">
 //    <label class="sb-label sb-padding-bottom-xs">Choose an option</label>
-//    <div class="sb-select__wrapper">
+//    <div class="sb-input-wrapper">
```
beerecca: Why change all this? I think the re-using of this class is just now a bit confusing. input-wrapper should be input__wrapper anyway (which i've noted down to fix) meaning it's part of the input component.

src/components/_dropdown.scss
```diff
@@ -0,0 +1,134 @@
...
+    width: 100%;
+    font-family: $sb-base-font-family;
+    font-size: sb-px2rems(13px);
+    border: 1px solid $sb-color-light-3;
+    border-radius: 0 0 sb-px2rems($sb-radius) sb-px2rems($sb-radius);
+    background-color: $sb-color-white;
+    cursor: pointer;
+    list-style: none;
+    padding: 0;
+    margin: -1px 0 0 0;
```
kdabo: The `ul` has a default margin. In order to reset it we can either write `margin: 0; margin: -1px;` or `margin: -1px 0 0 0;`

src/components/_dropdown.scss
```diff
@@ -0,0 +1,134 @@
...
+//
+// Basic dropdown
+//
+// .sb-dropdown--open - open dropdown
+//
+// Markup:
+//  <div class="sb-form-control">
+//    <label class="sb-label sb-padding-bottom-xs">Choose an option</label>
+//    <div class="sb-dropdown {{modifier_class}}">
+//      <div class="sb-input-wrapper">
```
kdabo: The `input-wrapper` is just wrapping the button of the dropdown. We have the same style for inputs that needs to be applied in many places. Namely: Select Component, Dropdown Button and Inputs. So either we duplicate all the styling or we use the same css classes.

Maybe there's a better way to extract these common styles somewhere?

src/components/_dropdown.scss
kdabo: Yeah we should put it everywhere, this can be part of another ticket (https://usabilla.atlassian.net/browse/STYLE-56)

src/components/_select.scss
kdabo: Currently the `sb-input-wrapper` modifies the styles of the `sb-input` and the `sb-icon`. Maybe we can extract the common styles for the input, dropdown, and the select in another class?

src/components/_dropdown.scss
beerecca: I tested it with just a margin-top -1px and it works - not sure why but the left margin seems fine, could you give it a go?
Shouldn't add it to all uls.

src/components/_dropdown.scss
```diff
@@ -0,0 +1,135 @@
...
+    &:hover {
+      background-color: $sb-color-light-4;
+    }
+
+    &:focus {
+      background-color: $sb-color-blue-4;
+      outline: none;
+    }
+
+    &:active {
```
dine: Not exactly an issue, but what do we think about grouping some pseudo selectors in one rule? For example:
```
&:focus,
&:active {
  background-color: $sb-color-blue-4;
  outline: none;
}
```

src/components/_dropdown.scss
```diff
@@ -0,0 +1,135 @@
...
+//
+// Markup:
+// <div class="sb-form-control {{modifier_class}}">
+//    <label class="sb-label sb-padding-bottom-xs">Choose an option</label>
+//    <div class="sb-dropdown">
+//      <div class="sb-dropdown__button sb-input-wrapper">
+//        <button class="sb-input">
+//          Choice 1
+//        </button>
+//        <svg class="sb-icon" viewBox="-5 -5 16 16"><path d="M0 0l3.5 4L7 0H0"></svg>
```
dine: This icon is going to be used in more than one component now, so I think we should add it to `_icons.scss`.

src/components/_select.scss
```diff
@@ -5,34 +5,28 @@
 // Markup:
 //  <div class="sb-form-control">
 //    <label class="sb-label sb-padding-bottom-xs">Choose an option</label>
-//    <div class="sb-select__wrapper">
-//      <select class="sb-select sb-input">
+//    <div class="sb-select sb-input-wrapper">
```
dine: Nice replacement! I liked the use of `.sb-input-wrapper` instead of `sb-select__wrapper`

src/components/_dropdown.scss
kdabo: `active` shouldn't have `outline: none`, only focus.

docs-builder/kss-assets/kss.scss
```diff
@@ -803,3 +803,7 @@ $kss-vertical-rhythm      : $kss-font-size * 1.5;
 
 .kss-color .kss-parameters__name { padding: .5rem; }
 .kss-color .kss-parameters__name code { visibility: hidden; }
+
+.sb-dropdown--open {
```
endritbajo: The `{{name}}` will place the class with the dot `class=".sb-dropdown--open"` instead of `class="sb-dropdown--open"`. Because of this `height: 155px` will override the stylabilla `.sb-dropdown-open` class height property instead of the kss property. In order to fix that we just need to change it to:

```css
.\.sb-dropdown--open {
  height: 155px;
}
```

src/components/_icons.scss
```diff
@@ -19,7 +19,9 @@
 // <svg class="sb-icon sb-icon--success" id="sb-icon-success" viewBox="0 0 16 16">
 //   <path d="M7,0 C3.1338125,0 0,3.1338125 0,7 C0,10.8661875 3.1338125,14 7,14 C10.8661875,14 14,10.8661875 14,7 C14,3.1338125 10.8661875,0 7,0 M5.57852066,10.5 L2.625,6.82666099 L3.93391119,5.5225724 L5.57896647,7.51873935 L10.733473,3.5 L11.375,4.11158433 L5.57852066,10.5"></path>
 // </svg>
-//
+//  <svg class="sb-icon sb-icon--arrow" id="sb-icon--arrow" viewBox="-5 -5 16 16">
```
beerecca: We don't have a class for sb-icon--arrow (and shouldn't)

beerecca: Is it possible to make this work with viewBox="0 0 16 16" to be consistent? No worries if not

kdabo: Why not?

And if there is not -5 -5, the arrow is not centered within it's tiny container

beerecca:
Nicceeeee :D

src/components/_dropdown.scss
```diff
@@ -0,0 +1,135 @@
...
+//        <li class="sb-dropdown__menu-item" tabindex="0" role="option">Choice 3</li>
+//      </ul>
+//    </div>
+//  </div>
+//
+// Styleguide dropdown
+
+.#{$ns}dropdown {
+  font-family: $sb-base-font-family;
+  font-size: sb-px2rems(13px);
```
endritbajo: Add `position: relative` here other wise the menu width will expand to the whole page!

PR #59, opened 2017-08-08
chore(accessibility): added pa11y

endritbajo:
## Changelog
- Added automated accessibility testing tool called pa11y

### Notes:
In order to test accessibility of a certain section you need to run the following commands:

```bash
# Need to have the server running
yarn dev
# Run the tests on section 9.2
yarn test:accessibility 9.2
```

If we want to test the accessibility of all the sections we just need to run:

```bash
# Need to have the server running
yarn dev
# Run the tests on all the sections
yarn test:accessibility
```

These will generate a terminal output like the following:

screen shot 2017-08-08 at 17 07 32

The header will be the url of the page that it is testing (in this case `http://localhost:8080/item-9-2.html`) and next it will print the outcome of that test.

The `yarn test:accessibility` command has 5 arguments:
```bash
yarn test:accessibility
   --url http://localhost:8080 # url of the dev server
   --dir ./docs # dir of static files generated by the styleguide
   --ignore 'notice,warning' # type of pa11y messages to ignore ('notice' or 'warning' or 'error')
   --concurrency 4 # number of pa11y tests to run in parallel
   9.1 # section of the styleguide where we want to run the accessibility test. If you want to test everything just skip this parameter
```

##### TODO: add documentation on how to test the accessibility with pa11y (already created a JIRA ticket https://usabilla.atlassian.net/browse/STYLE-55)


scripts/pa11y.js
```diff
@@ -0,0 +1,78 @@
...
+const chalk = require('chalk')
+const pa11y = require('pa11y')
+const parseArgs = require('minimist')
+
+const args = parseArgs(process.argv.slice(2))
+const url = args.host || 'http://localhost:8080/'
+const dir = args.dir || './docs'
+const concurrency = args.concurrency || 1
+const ignore = String(args.ignore || '')
+const section = String(args._[0] || '*')
```
endritbajo: We could improve the argument validation. Maybe we can improve it the next stylabilla day?

scripts/pa11y.js
```diff
@@ -0,0 +1,78 @@
+'use strict'
+
+const async = require('async')
+const glob = require('glob')
+const chalk = require('chalk')
+const pa11y = require('pa11y')
+const parseArgs = require('minimist')
+
+const args = parseArgs(process.argv.slice(2))
+const url = args.host || 'http://localhost:8080/'
```
endritbajo: Make this a mandatory argument?

scripts/pa11y.js
```diff
@@ -0,0 +1,78 @@
...
+const dir = args.dir || './docs'
```
endritbajo: Also this should be mandatory

scripts/pa11y.js
```diff
@@ -0,0 +1,78 @@
...
+const concurrency = args.concurrency || 1
+const ignore = String(args.ignore || '')
+const section = String(args._[0] || '*')
+
+const test = pa11y({
+  allowedStandards: ['WCAG2AA'],
+  ignore: ignore.split(',')
+})
```
endritbajo: Yeah good point, I wanted to do that but run out of time. I plan on adding it in the next iteration!

scripts/pa11y.js
```diff
@@ -0,0 +1,78 @@
...
+}
+
+const logNotice = (result) => {
+  log(chalk.cyan(' • Notice: ') + chalk.dim(result.context))
+  log('   ' + result.message)
+  log()
+}
+
+const logResults = (url, results) => {
+  logUrl(url)
```
endritbajo: Instead of not logging anything maybe we can log that the test for that url passed successfully? So the user can see that the test did actually run on that url

keepitterron: Yep, good point!

PR #60, opened 2017-08-10
Deploy docs

beerecca:
@endritbajo @kdabo 

There's quite a bit in this pr sorry, but needed to get everything working together. Have tested both `yarn build` and `yarn start` on an empty file, and after having run the other command, and hot reloading is still working and everything's coming out nicely. Could have created another "dist" folder for the docs but felt confusing for users of stylabilla, and not necessary as yarn build cleans the folder out and it's only ever compiled files anyway.

- Updated readme to point to the stylabilla sketch folder rather than the file which was broken
- Added logo inline (prevents pa11y complaining and nicer anyway), and style changes for this
- Moved the kss js, the necessary css and logo to be imported by kss.js (making the webpack and kss config much cleaner)
- Added command to deploy prod docs to s3 (originally was thinking github pages, but this is nicer, means you'll be able to access both `https://static.usabilla.com/stylabilla/v1.5.0/stylabilla.css` and `https://static.usabilla.com/stylabilla/v1.5.0/docs/index.html`
- Added webpack config to build the docs


webpack.config.prod.js
```diff
@@ -48,31 +58,99 @@ const config = {
         })
       },
       {
-        test: /\.(ttf|eot|woff|woff2)$/,
-        use: 'file-loader?name=assets/fonts/[name].[ext]&publicPath=/'
+        test: /\.(gif|png|jpg|svg|)$/,
```
katia-agna: is there a difference between `(gif|png|jpg|svg|)` and `(gif|png|jpg|svg)` without the pipe. Sometimes you put it sometimes not.

PR #62, opened 2017-08-11
feat(dropdown): add truncate text

kdabo:
Reviewers: @beerecca @endritbajo 

Added styles to truncate text if it doesn't fit in the dropdown box and fixed blue color styling for  focus. 

screen shot 2017-08-10 at 17 42 41


src/components/_dropdown.scss
```diff
@@ -59,12 +62,12 @@
   &__menu-item {
     padding: sb-px2rems(11px);
 
-    &:hover {
+    &:hover,
+    &:focus {
```
beerecca: Are you sure this is right? Did you talk to gal about this?

kdabo: yes, talked to Gal, and is all good.

PR #65, opened 2017-08-16
feat(toast): add toast component

beerecca:
@usabilla/mallard 
@usabilla/flamingo 
screen shot 2017-08-16 at 17 21 01


src/components/_toast.scss
```diff
@@ -0,0 +1,101 @@
...
+// Styleguide toast
+
+.#{$ns}toast {
+  display: flex;
+  align-items: stretch;
+  background: $sb-color-white;
+  position: fixed;
+  max-width: 40%;
+  top: sb-px2rems(10px);
+  right: sb-px2rems(10px);
```
keepitterron: Shouldn't thes be 11px?

src/components/_toast.scss
```diff
@@ -0,0 +1,101 @@
...
+  border: 1px solid $sb-color-light-3;
+  box-shadow: $sb-box-shadow;
+  opacity: 1;
+  z-index: 10;
```
keepitterron: Should we try and find a way to manage z-index on multiple levels? Like with a counter or something? It's kind of arbitrary like this and I'm pretty sure soon enough we're gonna need to bump it up

src/components/_toast.scss
```diff
@@ -0,0 +1,101 @@
...
+//  </div>
+//
+// Styleguide toast
+
+.#{$ns}toast {
+  display: flex;
+  align-items: stretch;
+  background: $sb-color-white;
+  position: fixed;
+  max-width: 40%;
```
keepitterron: What happens on mobile?

src/components/_toast.scss
```diff
@@ -0,0 +1,101 @@
...
+  top: sb-px2rems(10px);
```
endritbajo: What happens when we have multiple toasts one after another. Will they overlap?

src/components/_toast.scss
beerecca: It looks like shit 😅
Will fix. Have added a ticket to the styleguide project in jira to review the responsiveness of all our components.

src/components/_toast.scss
beerecca: Have added a ticket to the styleguide project. Next week we will have some time to tidy these things up on a more global perspective.

beerecca:
@keepitterron updated

src/components/_toast.scss
```diff
@@ -0,0 +1,107 @@
...
+//
+// Styleguide toast
+
+.#{$ns}toast {
+  display: flex;
+  align-items: stretch;
+  background: $sb-color-white;
+  position: fixed;
+  margin-left: sb-px2rems(11px);
+  top: sb-px2rems(11px);
```
endritbajo: What happens when we have multiple toasts? It looks like they will overlap

keepitterron: ah shit, you're right.
maybe we need a `toast-container` of some sort? and just append all the toasts there?

beerecca: Probably yes but this wasn't part of our requirements... tbh i think the design of the toasts is so large that they will look pretty average if multiple display, and I don't think we have multiple toasts anywhere at the moment? Would prefer to leave that requirement until we need it

PR #66, opened 2017-08-18
feat(dropdown): style for empty state and arrow rotate

kdabo:
**Reviewers:@endritbajo @beerecca @katia 

Changes:
- added style to rotate arrow on drop-down open
- added empty state style
- adjust the documentation style


src/components/_dropdown.scss
```diff
@@ -87,6 +93,11 @@
     .#{$ns}dropdown__menu {
       display: block;
     }
+
+    .#{$ns}icon--up {
+      transform: rotate(180deg);
+      top: 30%;
```
kdabo: not really happy with this solution, but if I don't add `top: 30%` the arrow is not centered.
do you guys have any idea why?

endritbajo: When you're adding the transform you're actually overriding the default `transform: translateY(-50%)`.
So you need to replace this with `transform: translateY(-50%) rotate(180deg)`

src/components/_dropdown.scss
```diff
@@ -12,9 +12,10 @@
 //        <button class="sb-input">
 //          Choice 1
 //        </button>
-//        <svg class="sb-icon" viewBox="-5 -5 16 16"><path d="M0 0l3.5 4L7 0H0"></svg>
+//        <svg class="sb-icon sb-icon--up" viewBox="-5 -5 16 16"><path d="M0 0l3.5 4L7 0H0"></svg>
```
endritbajo: I would name this `sb-icon--flip` so we can actually reuse it in the future in other components Ex. table headers, etc.

endritbajo: Also, I would this class dynamically only when we want it to rotate and not being by default there.

kdabo: Ok. I added it there for the sake of preview in the documentation.

src/components/_dropdown.scss
```diff
@@ -15,6 +15,7 @@
 //        <svg class="sb-icon" viewBox="-5 -5 16 16"><path d="M0 0l3.5 4L7 0H0"></svg>
 //      </div>
 //      <ul class="sb-dropdown__menu" role="listbox">
+//        <li class="sb-dropdown__menu-item sb-dropdown__menu-item--disabled" tabindex="0" role="option">No items to select</li>
```
endritbajo: Update the `tabindex="-1"` as we don't want it to be focusable

endritbajo: Maybe we can put it in a subsection as 'No options dropdown'

src/components/_dropdown.scss
```diff
@@ -61,7 +62,6 @@
   }
 
   &__menu-item {
-    padding: sb-px2rems(11px);
```
beerecca: Why no padding anymore?

src/components/_dropdown.scss
```diff
@@ -58,6 +59,7 @@
     margin: -1px 0 0 0;
     max-height: 300px;
     overflow-y: scroll;
+    z-index: 100;
```
endritbajo: Is this number for a reason or is it random? We should set it less than the `toast` z-index.

src/components/_dropdown.scss
```diff
@@ -138,4 +155,23 @@
 //  </div>
 //
 // Styleguide dropdown.validation
+
+// No options dropdown
```
endritbajo: We decided to remove the `No options` li menu item right?

PR #68, opened 2017-08-22
feat(input): BREAKING CHANGE: consolidated input styles

beerecca:

- Moved the label padding to the actual label, as it always needs it
- Combined label and input components as they should always be together
- Renamed `sb-input-wrapper` to `sb-form-control__input` and made it properly part of the form-control component which deals with validation and positioning
- Fixed bug that applied more right padding even when no icons were visible (30px comes from 7+7 padding + 16 width of icon)

Also once this is approved I will make appropriate changes to react-stylabilla.


src/components/_form-control.scss
```diff
@@ -42,58 +41,38 @@
...
-// Styleguide forms.input-wrapper
-
-.#{$ns}input-wrapper {
-  position: relative;
-
-  .#{$ns}input {
-    padding-right: sb-px2rems(23px);
-  }
+  &__input {
+    position: relative;
```
dine: ☝️ here I think we might need the padding-right, because we use the space for the arrow icon, no?

src/components/_input.scss
```diff
@@ -45,3 +43,8 @@
     color: $sb-color-light-3;
   }
 }
+
+.#{$ns}label {
+  display: inline-block;
+  margin-bottom: sb-px2rems(7px);
```
dine: nice👌

src/components/_form-control.scss
```diff
@@ -42,58 +41,38 @@
     .#{$ns}icon--success {
       display: inline;
     }
+
+    .#{$ns}input {
+      padding-right: sb-px2rems(30px);
```
dine: I like this idea, but I think we can't use the padding-right only here (check the `.sb-form-control__input` on  select and dropdown)

beerecca: ah damn you're right. ok will add back to the __input and put some documentation

beerecca: argh you're right. though i don't want to base the padding on the existence of the `sb-form-control__input` div, this makes the react component quite messy as it has to switch between rendering that div and not when the validation changes. so i think i'm just going to put the padding in the select and dropdown components.

beerecca:
@dine updated

PR #69, opened 2017-08-22
fix(upload): update button style, font-size and paddings

dine:
Closes [STYLE-59]
Reviewers: @beerecca @endritbajo @kdabo @galagmon 

## Changelog
- Changed inactive upload button to secondary style
- Updated font-size when active
- Updated button padding so it gets the same height as all input/upload fields.

## Screenshots
Before:

After:


src/settings/_box.scss
```diff
@@ -14,3 +14,8 @@ $sb-box-shadow-heavy: 0 1px 0 0 $sb-color-light-3, 0 2px 4px transparentize($sb-
   box-shadow: $sb-box-shadow;
   outline: none;
 }
+
+@mixin sb-focus-no-border() {
```
dine: @katia-agna you can use this mixin for your checkbox component 😊

PR #70, opened 2017-08-22
fix(buttons-group): add validation

dine:
Closes [STYLE-67]
Reviewers: @beerecca @endritbajo @kdabo @galagmon 

## Changelog
- Add example using validation on buttons-group
- Move buttons-group to `_buttons.scss` file

## Screenshots
Before:
screen shot 2017-08-22 at 12 35 45

After:
screen shot 2017-08-22 at 12 36 06


src/components/_buttons.scss
```diff
@@ -254,3 +254,94 @@ $sb-btn-padding-l: sb-px2rems(11px) sb-px2rems(23px);
 .#{$ns}btn--l {
   padding: $sb-btn-padding-l;
 }
+
+// Buttons Group
```
beerecca: Can we rename to Button Group?

beerecca: Sorry to say but I actually think a ButtonGroup is a separate component to Buttons and should be in a different file. a) the file is getting quite large b) button groups are actually closer to being a form element than a button c) and therefore should also have a label and validation, unlike a button. Assuming you are still changing the react component to support label and validation?

PR #71, opened 2017-08-22
refactor(z-index): consolidate z-index across stylabilla

kdabo:
Reviewers: @endritbajo @spirosikmd @katia @kdabo @NurahSadek @dine @beerecca 

As z-indexes were randomly defined, the team decided to review current z-indexes and outline general levels applicable to the entire application. Discussing with Gal, we came to a conclusion that it would be valid to have jumps of 10, leaving space for additional components' z-index that would be added.


kdabo:
I thought of it as a more general setting. I agree for `drop-down` or `toast`, but not really convinced for `sb-z-btngroup-2` as it would mean we can't reuse it for another purpose.  @beerecca

spirosikmd:
Also we should be careful of the type of commit message we chose, as we now create releases automatically from commit messages. In your case @kdabo this is not a new feature that should produce a new minor version imo, but a refactor? The `z-indexes` aren't available to the users of the stylabilla, right?

endritbajo:
I agree with @beerecca on the naming. I think we should find a solution which allows us to see the current `z-index` state of the components from a central place so we can fine tune it easily.

Even thought it's a bit old we could take inspiration from this: https://www.sitepoint.com/better-solution-managing-z-index-sass/

beerecca:
I think that sitepoint article goes a bit too far for me and isn't really necessary, but I do think we should name these semantically. Just noticed bootstrap does this also (https://getbootstrap.com/docs/4.0/layout/overview/#z-index). I'm not a huge fan of `sb-z-btngroup-2` either, but not because it means we can't reuse it, actually because it's not semantic enough. We have an infinite number of z-indexes we can use @kdabo so there's no issue or advantage with re-use. If they're all in the same file then there's no issue, we can just add new ones as we need them, specific for their use case. How about `sb-z-btn-active` and `sb-z-btngroup-hover`, rather than the numbers.

src/components/_button-group.scss
```diff
@@ -23,7 +23,7 @@
     display: inline-block;
     font-weight: 400;
     margin-right: -1px;
-    z-index: 1;
+    z-index: $sb-zindex-btn;
```
spirosikmd: I agree with @beerecca on naming and I would prefer something like `$sb-z-btn` instead of mentioning `zindex` everywhere. It is more readable for me!

src/components/_header.scss
```diff
@@ -25,7 +25,7 @@
   overflow: auto;
   background: $sb-color-white;
   display: flex;
-  z-index: 1;
+  z-index: $sb-z-header;
```
katia-agna: is it working the same? before it was the same weight as `$sb-z-btn` so it might change the behaviour somewhere

beerecca: Yes, please make sure the new z-indexes are tested against each other! :)

PR #77, opened 2017-08-22
feat(accessibility): add role alert on paragraph elements

katia-agna:


katia-agna:
I'll double check @spirosikmd  and @dine  but I add the feeling it was already add to other element

PR #78, opened 2017-08-22
fix(checkbox): apply different focus style

katia-agna:

## BEFORE
screen shot 2017-08-22 at 15 35 51

## AFTER
screen shot 2017-08-22 at 15 36 43


src/settings/_box.scss
```diff
@@ -16,6 +16,6 @@ $sb-box-shadow-heavy: 0 1px 0 0 $sb-color-light-3, 0 2px 4px transparentize($sb-
 }
 
 @mixin sb-focus-no-border() {
-  box-shadow: 0 0 4px 2px rgba(0, 165, 201, 0.25);
+  box-shadow: 0 0 4px 2px transparentize($sb-color-blue-2, 0.75);
```
dine: thanks 👍

src/components/_checkbox.scss
```diff
@@ -47,7 +47,7 @@
   }
 
   &:hover .#{$ns}checkbox__faux {
-    box-shadow: $sb-box-shadow;
+    @include sb-focus-no-border();
```
dine: I understood that this styles would be used just for :focus, should we update `_upload.scss` and apply to :hover too?

katia-agna: no, it is just for focus element which has the same background color has the focus border. in the case of upload.scss I don't see any problem.

PR #79, opened 2017-08-23
fix(accessibility): fix pa11y errors, except from contrast

dine:
Related to [STYLE-64]
Reviewers: @katia-agna @endritbajo @kdabo @beerecca 

## Changelog
- Fix all pa11y errors, except from contrast (which will be discussed with @galagmon)


src/components/_dropdown.scss
```diff
@@ -6,10 +6,10 @@
 //
 // Markup:
 //  <div class="sb-form-control">
-//    <label class="sb-label">Choose an option</label>
+//    <p class="sb-label">Choose an option</p>
 //    <div class="sb-dropdown {{modifier_class}}">
 //      <div class="sb-dropdown__button sb-form-control__input">
-//        <button class="sb-input">
+//        <button class="sb-input" title="Choose an option">
```
katia-agna: I'm not sure the title attribute on a button is the good thing to do. Especially if it's different than the visible "title" of the button.
More read about it : 
https://www.paciellogroup.com/blog/2012/01/html5-accessibility-chops-title-attribute-use-and-abuse/
https://www.w3.org/TR/wai-aria-practices/#button

src/components/_dropdown.scss
```diff
@@ -101,10 +101,10 @@
 //
 // Markup:
 //  <div class="sb-form-control">
-//    <label class="sb-label">Choose an option</label>
+//    <p class="sb-label">Choose an option</p>
 //    <div class="sb-dropdown {{modifier_class}}">
 //      <div class="sb-dropdown__button sb-form-control__input">
-//       <button class="sb-input" disabled>
+//       <button class="sb-input" title="Choose an option" disabled>
```
katia-agna: Same for all buttons type

src/components/_button.scss
```diff
@@ -4,13 +4,13 @@
 //
 // Markup:
 // <button class="sb-btn {{modifier_class}}">Button</button>
-// <button class="sb-btn {{modifier_class}}" aria-label="Star">
-//   <svg class="sb-icon" id="sb-icon-star" viewBox="0 0 16 16">
+// <button class="sb-btn {{modifier_class}}" title="Star">
```
katia-agna: We should keep the aria-label.

PR #83, opened 2017-08-24
fix(accessibility): accessibility fixes

dine:
Reviewers: @katia @endritbajo @kdabo @beerecca @spirosikmd 

## Changelog
- More accessibility fixes, pa11y errors and comments from @katia-agna on https://github.com/usabilla/stylabilla/pull/79
- Contrast colors were tested but not approved yet.


src/components/_button-group.scss
```diff
@@ -78,11 +78,11 @@
 // Markup:
 // <div class="sb-form-control {{modifier_class}}">
 //   <p class="sb-label">Button Group label</p>
-//   <div class="sb-btn-group">
-//     <button class="sb-btn">Button one</button>
-//     <button class="sb-btn">Button two</button>
-//     <button class="sb-btn sb-btn--active">Active</button>
-//     <button class="sb-btn" disabled>Disabled</button>
+//   <div class="sb-btn-group" role="radiogroup">
+//     <button role="radio" class="sb-btn">Button one</button>
```
katia-agna: Are they really radio button? I don't think this role is the most appropriate. I would not put any other type of rule than the implicit one (button).
Maybe you could use `role="region"` for the parent div. if you want to describe a "group" or define yourself the role with this new aria attribute https://www.w3.org/TR/wai-aria-1.1/#aria-roledescription

src/components/_button.scss
```diff
@@ -4,12 +4,12 @@
...
 // Markup:
 // <button class="sb-btn {{modifier_class}}">Button</button>
-// <button class="sb-btn {{modifier_class}}" title="Star">
+// <button class="sb-btn {{modifier_class}}" title="Star" aria-label="Star">
 //   <svg class="sb-icon" viewBox="0 0 16 16">
 //     <path d="M4.50044648,13.7056111 L8,11.5680972 L11.4995535,13.7056111 L10.5480786,9.71680978 L13.6623965,7.04906473 L9.57479916,6.7213638 L8,2.93509282 L6.42520085,6.7213638 L2.33760346,7.04906473 L5.45192146,9.71680977 L4.50044648,13.7056111 Z M8,12.6096808 L3.13994104,15.5781862 L4.4613166,10.0386768 L0.13625942,6.33380469 L5.81297339,5.878705 L8,0.620462738 L10.1870267,5.878705 L15.8637405,6.33380469 L11.5386834,10.0386768 L12.8600589,15.5781862 L8,12.6096808 Z"></path>
 //   </svg>
 // </button>
-// <button class="sb-btn {{modifier_class}}">
+// <button class="sb-btn {{modifier_class}}" title="Delete" aria-label="Delete">
```
beerecca: In this case it doesn't need `title="Delete" aria-label="Delete"` because there is an actual title on the button `Delete`.

beerecca: Let's add some comments to the pally script describing the rules we're ignoring and why we're ignoring them. And maybe check quickly to see if there's a way to ignore particular situations.

src/components/_dropdown.scss
```diff
@@ -9,7 +9,7 @@
 //    <p class="sb-label">Choose an option</p>
 //    <div class="sb-dropdown {{modifier_class}}">
 //      <div class="sb-dropdown__button sb-form-control__input">
-//        <button class="sb-input" title="Choose an option">
+//        <button class="sb-input" title="Choose an option" aria-label="Choose an option">
```
beerecca: Let's remove title and aria label, and add aria-labelledby instead.

src/components/_button.scss
beerecca: Looks like we may have to do a quick review of everywhere we're using `title`

PR #85, opened 2017-08-28
feat(dropdown): add multiple choice dropdown

katia-agna:
Reviewers: @endritbajo @spirosikmd @katia @kdabo @NurahSadek @dine @keepitterron @pietvanzoen @beerecca 
multiselect


src/components/_dropdown.scss
```diff
@@ -145,3 +145,62 @@
 //  </div>
 //
 // Styleguide dropdown.validation
+
+// Dropdown with multi select options
+//
+//
+// Markup:
+//  <div class="sb-form-control .sb-dropdown--open">
```
spirosikmd: We shouldn't have a `.` here!

src/components/_dropdown.scss
```diff
@@ -145,3 +145,62 @@
...
+//    <div class="sb-dropdown sb-dropdown--open">
+//      <div class="sb-dropdown__button sb-form-control__input">
+//        <button class="sb-input" title="Select options">
+//          Select options
+//        </button>
+//        <svg class="sb-icon" viewBox="-5 -5 16 16"><path d="M0 0l3.5 4L7 0H0"></svg>
+//      </div>
+//      <ul class="sb-dropdown__menu" role="listbox">
+//        <li class="sb-dropdown__menu-item" tabindex="0" role="option">
+//           <div class="sb-checkbox">
```
spirosikmd: Should we use two spaces tab here as in the rest?

src/components/_dropdown.scss
```diff
@@ -145,3 +145,62 @@
...
+//    <p class="sb-label">Select options</p>
+//    <div class="sb-dropdown sb-dropdown--open">
+//      <div class="sb-dropdown__button sb-form-control__input">
+//        <button class="sb-input" title="Select options">
+//          Select options
+//        </button>
+//        <svg class="sb-icon" viewBox="-5 -5 16 16"><path d="M0 0l3.5 4L7 0H0"></svg>
+//      </div>
+//      <ul class="sb-dropdown__menu" role="listbox">
+//        <li class="sb-dropdown__menu-item" tabindex="0" role="option">
```
endritbajo: Shouldn't we remove the `tabIndex` from the li and just leave the checkbox input?

src/components/_dropdown.scss
```diff
@@ -145,3 +145,62 @@
...
+//                 <svg class="sb-checkbox__checkmark" viewBox="0 0 8 7">
+//                   <path d="M7.68025,0 C5.067,1.97775 2.56075,4.6235 2.56075,4.6235 L0.7545,3 L0,3.7685 C0.74225,4.46625 2.46225,6.30925 3.044,7 C4.6815,4.26425 6.38375,2.224 8,0.3585 L7.68025,0 Z"></path>
+//                 </svg>
+//               </span>
+//               <span class="sb-margin-left-xs">Choice 1</span>
+//             </label>
+//           </div>
+//        </li>
+//        <li class="sb-dropdown__menu-item" tabindex="0" role="option">
+//           <div class="sb-checkbox">
```
endritbajo: Also we should be able to tick the checkbox when we click anywhere on row right? Right now it can be ticked only when clicking on the label or the square

spirosikmd:
I notice the mouse cursor becomes `default` when mouse over the checkbox and `pointer` when mouse over the item. Should this be consistent?

src/components/_dropdown.scss
katia-agna: it looks strange I know but it the only way to trigger the "hack" @kdabo made last time for dropdown open. 
```
.\.sb-dropdown--open {
  height: 200px;
}
```
Maybe we did that to not confuse a real classe. 
And I don't want to have a "close" state and an "open" state because it messes up the id checkbox so that the solution we found to set the class directly and not with a modifier.

src/components/_dropdown.scss
katia-agna: good point. But I'll set -1 like that it's still focusable with the Javascript (I think you'll need that for the react component)

endritbajo: No we don't need that for the react component as we are focusing directly on the checkbox input

katia-agna: but we still want to affect the css style of the row. like it is selected no ?

katia-agna:
@spirosikmd I agree with you but the checkbox component doesn't have this cursor. Should I set it in the checkbox component directly instead of doing that in the dropdown?

spirosikmd:
Well I guess it is up to design to decide? What is the common thing out there? Is a pointer cursor used on checkboxes?

katia-agna:
Not by default: https://www.w3schools.com/Html/tryit.asp?filename=tryhtml_checkbox
So I will not implement it. Since we also decided click on the `<li>` instead of the checkbox

katia-agna:
Are you ok with this presentation? 
(Don't pay attention to the background tile height which still needs to be fixed)

screen shot 2017-08-29 at 14 49 28

src/components/_dropdown.scss
```diff
@@ -145,3 +153,123 @@
 //  </div>
 //
 // Styleguide dropdown.validation
+
+// Dropdown with multi options
```
spirosikmd: I would write that as `Dropdown with multiple selection`

src/components/_dropdown.scss
```diff
@@ -145,3 +153,123 @@
...
+//           </label>
+//         </div>
+//       </li>
+//     </ul>
+//   </div>
+// </div>
+//
+// Styleguide dropdown.multiselect
+
+// Dropdown with multi option selected
```
spirosikmd: And this one as `Dropdown with multiple selection and selected item`?

beerecca:
@katia-agna just saw endrit approved, i'm reviewing now if you can wait to merge it pretty please

src/components/_dropdown.scss
```diff
@@ -80,6 +80,14 @@
     &:focus {
       outline: none;
     }
+
+    &--selected {
+      background-color: $sb-color-light-4;
+    }
+
+    .#{$ns}checkbox {
+      pointer-events: none;
```
beerecca: This means the js will have to set the checkbox value to checked itself? Why is this needed?

src/components/_dropdown.scss
```diff
@@ -146,3 +154,123 @@
...
+// Dropdown with multi options
+//
+// Default styling
+//
+// Markup:
+// <div class="sb-form-control">
+//   <p class="sb-label">Select options</p>
+//   <div class="sb-dropdown sb-dropdown--open">
+//     <div class="sb-dropdown__button sb-form-control__input">
+//       <button class="sb-input" title="Select options">
```
beerecca: This should be consistent with the normal dropdown (which currently has an aria-label). Don't know which is best but should be the same :)

src/components/_dropdown.scss
```diff
@@ -146,3 +154,123 @@
...
+//       </li>
+//     </ul>
+//   </div>
+// </div>
+//
+// Styleguide dropdown.multiselect
+
+// Dropdown with multiple selection and selected item
+//
+// .sb-dropdown__menu-item--selected
```
beerecca: This doesn't need to be a separate component. As far as the documentation goes I think we just need the one example with an item that is selected. And we don't need this comment (line 220), this just comes up as the documentation (I think you were trying to add a variation, the syntax for this is different, as per line 135) but if we just have the one version it's not needed anyway.

beerecca:
Can we also fix the height of the background of the dropdown in the documentation?

beerecca:
Oh and I agree with dine about the background versus the checkbox - if we're not going to support in the css that's fine but we should add documentation as to what needs to be done with js (similar to how the upload component documents it.

src/components/_dropdown.scss
katia-agna: for avoiding to check the checkbox by the user. we thought it was better to disable this behavior and managing in the javascript. instead of letting the user able to click on the checkbox but that doesn't focus the row.

beerecca: ok, let's document that then

beerecca: actually hold on - why don't we remove this and just document the additional js that is needed to focus the row? otherwise the js has to do both

katia-agna:
screen shot 2017-08-30 at 15 09 53

docs-builder/kss-assets/kss.scss
```diff
@@ -820,12 +929,15 @@ $kss-vertical-rhythm      : $kss-font-size * 1.5;
...
 }
 
-//Styles to override stylabilla for documentation
+.kss-modifier__example .sb-toast {
+  position: relative;
+}
 
-.\.sb-dropdown--open {
-  height: 200px;
+//Styles to override stylabilla for documentation
```
beerecca: This line should be above the `sb-toast` styles line 932

src/components/_dropdown.scss
```diff
@@ -98,6 +106,67 @@
   }
 }
 
+// Dropdown with multi options
+//
+//JavaScript is required for:
+// - Check/Unchecked the checkbox element on click handler of the row
```
beerecca: Should be `Checking/unchecking the`...

src/components/_dropdown.scss
```diff
@@ -98,6 +106,67 @@
...
+//             <span class="sb-checkbox__faux">
+//               <svg class="sb-checkbox__checkmark" viewBox="0 0 8 7">
+//                 <path d="M7.68025,0 C5.067,1.97775 2.56075,4.6235 2.56075,4.6235 L0.7545,3 L0,3.7685 C0.74225,4.46625 2.46225,6.30925 3.044,7 C4.6815,4.26425 6.38375,2.224 8,0.3585 L7.68025,0 Z"></path>
+//               </svg>
+//             </span>
+//             <span class="sb-margin-left-xs">Choice 2</span>
+//           </label>
+//         </div>
+//       </li>
+//       <li class="sb-dropdown__menu-item" tabindex="-1" role="option">
```
spirosikmd: The `tabindex` here I think should be `0`. And I guess if we don't want focus on the checkboxes, then all checkboxes should have `tabindex="-1"`?

src/components/_dropdown.scss
```diff
@@ -98,6 +106,67 @@
   }
 }
 
+// Dropdown with multi options
```
spirosikmd: I am still not fun of this.. Maybe `Dropdown with multiple item selection`?

PR #86, opened 2017-08-29
feat(input): input with static container - display information

katia-agna:
input info_real-one

### DIFF
--- a/src/components/_input.scss
+++ b/src/components/_input.scss
@@ -48,3 +48,38 @@
   display: inline-block;
   margin-bottom: sb-px2rems(7px);
 }
+
+// Input with info
+//
+// Markup:
+// <label class="sb-label" for="input-id-1">Input Label</label>
+// <div class="sb-input sb-input-with-info" tabindex="-1">
+//  <span class="sb-info">20 items</span>
+//  <input type="text" aria-autocomplete="list" class="sb-input-info" id="input-id-1" placeholder="Placeholder Text">
+// </div>
+//
+// Styleguide input.info
+
+.#{$ns}input-with-info {
+  display: flex;
+
+  .#{$ns}info {
+    color: $sb-color-blue-1;
+    padding-right: sb-px2rems(11px);
+    font-weight: 400;
+    font-size: sb-px2rems(13px);
+  }
+
+  .#{$ns}input-info {
+    flex: 1;
+    border: none;
+
+    &::placeholder {
+      color: $sb-color-light-1;
+    }
+
+    &:focus {
+      @include sb-no-focus-style();
+    }
+  }
+}
--- a/src/components/_input.scss
+++ b/src/components/_input.scss
@@ -10,15 +10,16 @@
 //
 // Styleguide input
 
+$padding-input-element: sb-px2rems(11px);
+
 .#{$ns}input {
+  @include sb-font-text-small();
+
   width: 100%;
   border-radius: sb-px2rems($sb-radius);
   border: 1px solid $sb-color-light-3;
   background-color: $sb-color-white;
-  padding: sb-px2rems(11px);
-  font-family: $sb-base-font-family;
-  font-weight: 400;
-  font-size: sb-px2rems(13px);
+  padding: $padding-input-element;
   color: $sb-color-dark-3;
 
   &::placeholder {
@@ -51,35 +52,42 @@
 
 // Input with info
 //
+// .sb-input__wrapper--hide-info - hide info element
+//
 // Markup:
-// <label class="sb-label" for="input-id-1">Input Label</label>
-// <div class="sb-input sb-input-with-info" tabindex="-1">
-//  <span class="sb-info">20 items</span>
-//  <input type="text" aria-autocomplete="list" class="sb-input-info" id="input-id-1" placeholder="Placeholder Text">
+// <label class="sb-label" for="input-id-2">Input Label</label>
+// <div class="sb-input__wrapper {{modifier_class}}">
+//  <span class="sb-input__info">20 items</span>
+//  <input type="text" class="sb-input" id="input-id-2" placeholder="Placeholder Text">
 // </div>
 //
 // Styleguide input.info
 
-.#{$ns}input-with-info {
-  display: flex;
+.#{$ns}input__wrapper {
+  position: relative;
 
-  .#{$ns}info {
+  .#{$ns}input__info {
+    @include sb-font-text-small();
+
     color: $sb-color-blue-1;
-    padding-right: sb-px2rems(11px);
-    font-weight: 400;
-    font-size: sb-px2rems(13px);
+    position: absolute;
+    bottom: $padding-input-element;
+    left: $padding-input-element;
+    line-height: 1.4;
   }
 
-  .#{$ns}input-info {
-    flex: 1;
-    border: none;
+  .#{$ns}input {
+    padding-left: sb-px2rems(53.19px) + (2 * $padding-input-element);
+  }
 
-    &::placeholder {
-      color: $sb-color-light-1;
+  &--hide-info {
+    .#{$ns}input__info {
+      display: none;
     }
 
-    &:focus {
-      @include sb-no-focus-style();
+    .#{$ns}input {
+      padding-left: $padding-input-element;
     }
   }
 }
+
--- a/src/settings/_box.scss
+++ b/src/settings/_box.scss
@@ -19,3 +19,9 @@ $sb-box-shadow-heavy: 0 1px 0 0 $sb-color-light-3, 0 2px 4px transparentize($sb-
   box-shadow: 0 0 4px 2px transparentize($sb-color-blue-2, 0.75);
   outline: none;
 }
+
+@mixin sb-no-focus-style() {
+  box-shadow: none;
+  outline: none;
+  border: none;
+}
--- a/src/settings/_box.scss
+++ b/src/settings/_box.scss
@@ -19,9 +19,3 @@ $sb-box-shadow-heavy: 0 1px 0 0 $sb-color-light-3, 0 2px 4px transparentize($sb-
   box-shadow: 0 0 4px 2px transparentize($sb-color-blue-2, 0.75);
   outline: none;
 }
-
-@mixin sb-no-focus-style() {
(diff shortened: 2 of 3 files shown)


katia-agna:
I wonder if I should do a state "hide info" by css or if we can play with two different component.

src/components/_input.scss
```diff
@@ -48,3 +48,38 @@
...
+// Markup:
+// <label class="sb-label" for="input-id-1">Input Label</label>
+// <div class="sb-input sb-input-with-info" tabindex="-1">
+//  <span class="sb-info">20 items</span>
+//  <input type="text" aria-autocomplete="list" class="sb-input-info" id="input-id-1" placeholder="Placeholder Text">
+// </div>
+//
+// Styleguide input.info
+
+.#{$ns}input-with-info {
```
dine: This looks like a modifier, so `.#{$ns}input--with-info` here

src/components/_input.scss
```diff
@@ -48,3 +48,38 @@
...
+  display: flex;
+
+  .#{$ns}info {
+    color: $sb-color-blue-1;
+    padding-right: sb-px2rems(11px);
+    font-weight: 400;
+    font-size: sb-px2rems(13px);
+  }
+
+  .#{$ns}input-info {
```
dine: And here it's a element, so I'd use `.#{$ns}input__info`

src/components/_input.scss
```diff
@@ -48,3 +48,38 @@
...
+    flex: 1;
+    border: none;
+
+    &::placeholder {
+      color: $sb-color-light-1;
+    }
+
+    &:focus {
+      @include sb-no-focus-style();
```
dine: Why no focus?

src/components/_input.scss
katia-agna: yeah that the concern my question at the top of this PR. Since the component is really different I'm not sure we can just play with the modifier. 
If you look the html markup it's not "just" a <span> I add inside the input. You know what I mean? Should I still work only with a modifier ?

src/components/_input.scss
katia-agna: because the style focus is handling by the parent element.

src/components/_input.scss
```diff
@@ -48,3 +48,38 @@
   display: inline-block;
   margin-bottom: sb-px2rems(7px);
 }
+
+// Input with info
+//
+// Markup:
+// <label class="sb-label" for="input-id-1">Input Label</label>
+// <div class="sb-input sb-input-with-info" tabindex="-1">
+//  <span class="sb-info">20 items</span>
```
spirosikmd: Should we use two spaces indentation here?

src/components/_input.scss
spirosikmd: I agree with @dine on this one. For me `sb-input--with-info` makes more sense in this case.

src/components/_input.scss
```diff
@@ -48,3 +48,38 @@
   display: inline-block;
   margin-bottom: sb-px2rems(7px);
 }
+
+// Input with info
+//
+// Markup:
+// <label class="sb-label" for="input-id-1">Input Label</label>
+// <div class="sb-input sb-input-with-info" tabindex="-1">
```
beerecca: To be consistent with the normal input, the `sb-input` class should be on the actual input element. This div is not an input, it's a wrapper. Therefore I would call it `sb-input__wrapper`. It will be useful if we have other things we will want to put in an input, like a unit indicator or something.

Also as dine mentioned, it should be `sb-input--with-info`. However I actually don't think we need this class, we can use `sb-input__wrapper` instead. If the only style is display flex then this is generic enough. Although I agree with spyros that we should try relative/absolute positioning if possible.

Also why do we need tabindex="-1"? A normal div wrapper shouldn't get focus anyway.

src/components/_input.scss
beerecca: This name is too generic. I think the ideal markup would be:
```
<div class="sb-input__wrapper">
<span class="sb-input__info">20 items</span>
<input type="text" class="sb-input" id="input-id-2">
</div>
```

src/components/_input.scss
```diff
@@ -48,3 +48,38 @@
...
+//  <span class="sb-info">20 items</span>
+//  <input type="text" aria-autocomplete="list" class="sb-input-info" id="input-id-1" placeholder="Placeholder Text">
```
beerecca: Why `aria-autocomplete="list"`?

src/components/_input.scss
```diff
@@ -48,3 +48,38 @@
...
+  .#{$ns}info {
+    color: $sb-color-blue-1;
+    padding-right: sb-px2rems(11px);
+    font-weight: 400;
+    font-size: sb-px2rems(13px);
+  }
+
+  .#{$ns}input-info {
+    flex: 1;
+    border: none;
```
beerecca: Yea we shouldn't be overriding borders... My suggestion above prevents this.

src/components/_input.scss
```diff
@@ -48,3 +48,38 @@
...
+
+    &::placeholder {
```
beerecca: Why is this placeholder a different color?

src/components/_input.scss
beerecca: But it shouldn't need to be handled by the parent element? If we leave `sb-input` the way it is and just add a wrapper, I think everything becomes simpler.

katia-agna:
hey, @spirosikmd @dine @beerecca I changed completely the structure of the component. Could you check again? Also, I still don't know if I have to work with a modifier_class or like that is enough for the implementation?

src/components/_input.scss
```diff
@@ -48,3 +54,22 @@
...
+// Input with info
+//
+// Markup:
+// <label class="sb-label" for="input-id-2">Input Label</label>
+// <span class="sb-info">20 items</span>
+// <input type="text" class="sb-input sb-input--with-info" id="input-id-2" placeholder="Placeholder Text">
+//
+// Styleguide input.info
+
+.#{$ns}info {
```
spirosikmd: I agree with @beerecca that this is too generic. Should be something like `sb-input__info`?

src/settings/_box.scss
```diff
@@ -19,3 +19,9 @@ $sb-box-shadow-heavy: 0 1px 0 0 $sb-color-light-3, 0 2px 4px transparentize($sb-
   box-shadow: 0 0 4px 2px transparentize($sb-color-blue-2, 0.75);
   outline: none;
 }
+
+@mixin sb-no-focus-style() {
```
spirosikmd: This is not used anymore, right? Do we need it?

src/components/_input.scss
```diff
@@ -48,3 +54,22 @@
...
+  color: $sb-color-blue-1;
+  font-weight: 400;
```
spirosikmd: Don't we have classes for text? Feels repetitive to specify `font-weight`, `font-size`, and `line-height`?

src/components/_input.scss
```diff
@@ -48,3 +54,22 @@
...
+  font-size: sb-px2rems(13px);
+  position: absolute;
```
beerecca: I don't think this is working. If it's position absolute then it is positioning itself relative to it's parent, which is probably a kss wrapper which is why it probably looks like it works. But I think you are gonna need a parent as part of the component :)

spirosikmd: Ooo yes I totally missed that...

katia-agna: ho boy ! ok I'll update that. thanks @beerecca

src/components/_input.scss
katia-agna: I didn't know it was user somewhere else. Will try to find this method or create one ! 👍

src/settings/_box.scss
katia-agna: no indeed we don't need that anymore

src/components/_input.scss
```diff
@@ -48,3 +49,45 @@
...
+
+    color: $sb-color-blue-1;
+    position: absolute;
+    bottom: $padding-input-element;
+    left: $padding-input-element;
+    line-height: 1.4;
+  }
+
+  .#{$ns}input {
+    padding-left: sb-px2rems(53.19px) + (2 * $padding-input-element);
```
dine: This number will only work if we have a info with size similar to `20 items`. If it is bigger it will override and smaller would have too much space after. We need another solution that makes this more dynamic / translatable.

src/components/_input.scss
```diff
@@ -48,3 +49,45 @@
...
+// Markup:
+// <label class="sb-label" for="input-id-2">Input Label</label>
+// <div class="sb-input__wrapper {{modifier_class}}">
+//  <span class="sb-input__info">20 items</span>
+//  <input type="text" class="sb-input" id="input-id-2" placeholder="Placeholder Text">
+// </div>
+//
+// Styleguide input.info
+
+.#{$ns}input__wrapper {
```
endritbajo: I think the name `input_wrapper` is bit confusing as the class is essentially a wrapper for an input with some infos (of length < `53.19px`) placed on the left and the name should reflect that. Therefore I think this may be better off as a component on it's own, something like `tag-input` component.

katia-agna: Well, we wanted to that later by extracting it but now it also doesn't look like a "tag" so it might be confusing to call it like that no ? 
for the naming of the `input_wrapper`, would `input__tag__wrapper` work better ?

src/components/_input.scss
katia-agna: yes I know. We had a long conversation with All flamingo members + Gal about it. 
I tried an other solution which was more dynamic and responsive but the work we had to do in javascript was too big. So this is the solution we came up with. And yes it will stay the same with it's 1 item or 99 items (no other word than item) for now. We assume it'll not be more and the Javascript could resize it in case of. But do you have an other idea ?

src/components/_input.scss
beerecca: The idea was to make it a generic wrapper that you could use to also put right aligned dimension text or something else in the future. I would prefer to keep it as it is. The only style it actually has is position relative.

katia-agna:
After discussing with Design and other developers. We decide to not implement this solution.

PR #87, opened 2017-08-29
fix(dropdown): change highlight style

katia-agna:
Reviewers : @usabilla/flamingo @usabilla/mallard


src/components/_dropdown.scss
```diff
@@ -53,14 +53,16 @@
...
     list-style: none;
     padding: 0;
-    margin: -1px 0 0 0;
+    margin: 0;
     max-height: 300px;
     overflow-y: scroll;
     z-index: 9;
+    border-bottom: 1px solid $sb-color-light-3;
+    border-right: 1px solid $sb-color-light-3;
+    border-left: 1px solid $sb-color-light-3;
```
keepitterron: I prefer
```
border: 1px solid $sb-color-light-3;
border-top: 0;
```
less code and it makes it explicit that we do not want a border top.
what do you think?

katia-agna: Yes, that makes sense! thanks

PR #88, opened 2017-08-29
docs(label): set html label when it is a label

katia-agna:
Reviewers : @usabilla/flamingo @usabilla/mallard


spirosikmd:
Imo, changes that affect documentation code should not trigger a release (so no `breaking change`, `feat`, or `fix`). For example this PR changes only code that used for documentation. Not actual css code. We don't really want to trigger a new release in this case, right? What do you think?

katia-agna:
I talked with @dine and we realized that this modification was made because of accessibility concern: https://www.w3.org/TR/WCAG20-TECHS/H44
this PR is no longer needed.

PR #89, opened 2017-08-31
docs(dropdown): Add multi select dropdown with search element

katia-agna:
screen shot 2017-09-04 at 14 00 06


src/components/_dropdown.scss
```diff
@@ -167,6 +167,70 @@
 //
 // Styleguide dropdown.multiselect
 
+// Dropdown with multiple item selection and search input
+//
+//JavaScript is required for:
```
kdabo: is this necessary? I don't think we added this kind of documentation for any component.

katia-agna: yes in the upload component we did. I followed the recommendation of @beerecca

src/components/_dropdown.scss
```diff
@@ -167,6 +167,70 @@
...
+// - Checking/Unchecking the checkbox element on click handler of the row
+// - Update the content of the dropdown depending of the input in the  input search
```
spirosikmd: Extra space in `the  input`?

src/components/_dropdown.scss
```diff
@@ -167,6 +167,70 @@
...
+// - Update the amount of element selected in the input search
```
spirosikmd: Maybe we should write this as `Update the amount of selected elements in the input search`?

src/components/_dropdown.scss
```diff
@@ -167,6 +167,70 @@
...
+//
+// Markup:
+// <div class="sb-form-control">
+//   <p class="sb-label">Select options</p>
+//   <div class="sb-dropdown sb-dropdown--open">
+//     <div class="sb-dropdown__button sb-form-control__input" role="search">
+//       <div class="sb-input__wrapper {{modifier_class}}">
+//         <span class="sb-input__info" id="input-info">20 items</span>
+//         <input type="text" class="sb-input" id="input-id-3" placeholder="Placeholder Text" aria-labeledby="input-info">
+//       </div>
```
endritbajo: This should be a button right?

src/components/_dropdown.scss
```diff
@@ -167,6 +167,70 @@
...
+//       <svg class="sb-icon" viewBox="-5 -5 16 16"><path d="M0 0l3.5 4L7 0H0"></svg>
+//     </div>
+//     <ul class="sb-dropdown__menu" role="listbox">
+//       <li class="sb-dropdown__menu-item" tabindex="0" role="option">
+//         <div class="sb-checkbox">
```
endritbajo: I don't see the search input in the dropdown menu 🤔

src/components/_dropdown.scss
```diff
@@ -82,12 +82,23 @@
...
-      background-color: $sb-color-light-4;
+      background-color: $sb-color-blue-4;
     }
 
     .#{$ns}checkbox {
       pointer-events: none;
     }
+
+    &__search {
+      padding: 0;
```
endritbajo: This seems to be superfluous, is there any reason why we put it there?

src/components/_dropdown.scss
```diff
@@ -168,6 +180,59 @@
...
+//   <p class="sb-label">Select options</p>
+//   <div class="sb-dropdown sb-dropdown--open">
+//     <div class="sb-dropdown__button sb-form-control__input">
+//       <button class="sb-input" title="Select options" aria-label="One item selected">
+//        One item selected
+//       </button>
+//       <svg class="sb-icon" viewBox="-5 -5 16 16"><path d="M0 0l3.5 4L7 0H0"></svg>
+//     </div>
+//     <ul class="sb-dropdown__menu" role="listbox">
+//       <li class="sb-dropdown__menu-item__search" role="search">
```
endritbajo: It would be better to name this class `sb-dropdown__menu-search` according to bem rules.

src/components/_dropdown.scss
katia-agna: Otherwise, every items row has a padding I don't get why you think it's superflous

endritbajo: Because with the suggested changes in my other comment, the `dropdown_search` is not a `menu-item` anymore, therefore there's no default padding applied to it by default so there's no need to override the `padding: 0`

src/components/_dropdown.scss
```diff
@@ -168,6 +180,59 @@
 //
 // Styleguide dropdown.multiselect
 
+// Dropdown with multiple item selection and search input
+//
+// JavaScript is required for:
+// - Checking/Unchecking the checkbox element on click handler of the row
+// - Updating the content of the dropdown depending of the input in the search element
+// - Updating the amount of selected element in the button search
```
endritbajo: What do think of `Updating the button text with the number of selected items`?

src/components/_dropdown.scss
```diff
@@ -82,14 +82,25 @@
...
+      background-color: $sb-color-blue-4;
     }
 
     .#{$ns}checkbox {
       pointer-events: none;
     }
   }
 
+  &__search {
+    padding: 0;
```
spirosikmd: This is not needed anymore!

src/components/_dropdown.scss
```diff
@@ -82,14 +82,25 @@
...
+
+    .#{$ns}input {
+      border-radius: 0;
+      box-shadow: none;
+      border: 0;
+      border-bottom: 1px solid $sb-color-light-3;
```
spirosikmd: We are reusing `1px solid $sb-color-light-3` in multiple places? Should we define a variable for this, or is it too much? @dine @endritbajo?

src/components/_dropdown.scss
```diff
@@ -168,6 +180,59 @@
 //
 // Styleguide dropdown.multiselect
 
+// Dropdown with multiple item selection and search input
+//
+// JavaScript is required for:
+// - Checking/Unchecking the checkbox element on click handler of the row
```
endritbajo: Maybe `Checking/Unchecking the checkbox on row click`?

src/components/_dropdown.scss
```diff
@@ -168,6 +180,59 @@
...
+// - Updating the content of the dropdown depending of the input in the search element
```
endritbajo: What about `Filtering the menu items based on the search value`

PR #90, opened 2017-08-31
docs(dropdown): Add aria-expanded property

katia-agna:


spirosikmd:
We should not use `feat` for this change I think? As you just update the docs, we don't want to update the minor version. Maybe just `doc(dropdown)`. We should also agree on scopes and validate them manually on commit level for consistency and for valuable changelogs?

PR #91, opened 2017-08-31
fix(header): make header responsive and match designs

beerecca:
BREAKING CHANGE: header markup has changed, so will need to update react-stylabilla

@dine @usabilla/flamingo @galagmon 

Note that I have added the `font-weight: 400` to override the browser overrides of the font-weight we set on `html` :|
Still need to do [STYLE-49](https://usabilla.atlassian.net/browse/STYLE-49) to review typography overall, but we needed this fix for the button creator.
screen shot 2017-08-31 at 17 17 12
screen shot 2017-08-31 at 17 17 32


galagmon:
Great! the only thing now is to align the sub title to the same base as the title as we discussed. And I can take a look at it tomorrow

beerecca:
Ah yes, about that - I tried it with baseline alignment but it doesn't work so well if you have a long title:
screen shot 2017-09-01 at 09 55 56

I checked the existing header in the app and it's centred, which I feel is much safer.

src/components/_header.scss
```diff
@@ -20,34 +22,26 @@
...
-
-  &__info {
-    line-height: sb-px2rems(37px);
-  }
-
-  &__btn {
-    float: right;
-    margin: sb-px2rems(7px) sb-px2rems(16px);
+  &__content {
+    display: flex;
```
dine: Does this works well on <IE10? I was using the same and Endrit mentioned it on this same component, check https://github.com/usabilla/stylabilla/pull/31#pullrequestreview-52030794

beerecca: No, looks like shit on IE9, but it works on IE10 and 11, and the header is only used in the app, so I think that's fine. Our browser support for the app is IE11+ I think? Not IE9 in any case.

src/components/_text.scss
```diff
@@ -19,37 +19,44 @@
...
 .#{$ns}title--s {
   font-size: sb-px2rems(19px);
   line-height: sb-px2rems(19px);
+  font-weight: 400;
 }
 
 .#{$ns}text {
   font-size: sb-px2rems(16px);
   line-height: sb-px2rems(23px);
+  font-weight: 400;
```
spirosikmd: Isn't font-weight: 400 for html? Why do we need reset it here to 400?

beerecca:
Updated as per discussion with @galagmon. Note that I didn't have to add any top padding by the way Gal, the reason it looked a little off centre was because the baseline accounts for descenders, for instance: 
screen shot 2017-09-03 at 18 23 26

src/components/_button-group.scss
```diff
@@ -3,11 +3,11 @@
 // Group a series of buttons together on a single line.
 //
 // Markup:
-// <div class="sb-btn-group" role="radiogroup">
-//   <button class="sb-btn" role="radio" aria-checked="false">Button one</button>
-//   <button class="sb-btn" role="radio" aria-checked="false">Button two</button>
-//   <button class="sb-btn sb-btn--active" role="radio" aria-checked="true">Active</button>
-//   <button class="sb-btn" disabled role="radio" aria-checked="false">Disabled</button>
+// <div class="sb-btn-group">
```
spirosikmd: Why did we remove `role`s and `aria-label`s? Is this a wrong conflict resolution?

src/components/_button-group.scss
```diff
@@ -74,16 +74,16 @@
...
 // Markup:
 // <div class="sb-form-control {{modifier_class}}">
-//   <p class="sb-label" id="label-btn-group">Button Group label</p>
-//   <div class="sb-btn-group" role="group" aria-labelledby="label-btn-group">
-//     <button class="sb-btn" role="checkbox" aria-checked="false">Button one</button>
-//     <button class="sb-btn" role="checkbox" aria-checked="false">Button two</button>
-//     <button class="sb-btn sb-btn--active" role="checkbox" aria-checked="true">Active</button>
-//     <button class="sb-btn" disabled role="checkbox" aria-checked="false">Disabled</button>
+//   <p class="sb-label">Button Group label</p>
+//   <div class="sb-btn-group" aria-roledescription="button group">
```
spirosikmd: Some for this code here?

PR #95, opened 2017-09-07
feat(dropdown): add no results message

spirosikmd:
screen shot 2017-09-11 at 09 50 13


spirosikmd:
Accessibility wise I am not sure if the `aria-label` is enough in the list element. Maybe @katia-agna has an idea about this one?

src/components/_dropdown.scss
```diff
@@ -229,6 +234,31 @@
...
 // Styleguide dropdown.multiselect.search
 
+// Dropdown with no search results
+//
+// Markup:
+// <div class="sb-form-control">
+//   <p class="sb-label">Select options</p>
+//   <div class="sb-dropdown sb-dropdown--open">
+//     <div class="sb-dropdown__button sb-form-control__input">
+//       <button class="sb-input" title="Select options" aria-label="One item selected">
```
katia-agna: it looks a little strange to have this button with this information "One item selected" no ?

PR #96, opened 2017-09-07
fix(header): paragraph optional and align baseline

dine:
@spirosikmd @beerecca @endritbajo @kdabo


src/components/_header.scss
```diff
@@ -12,8 +12,8 @@
 //   <div class="sb-header__content">
 //     <h1 class="sb-header__title sb-title">Header title</h1>
 //     <p class="sb-header__info sb-text">Info regarding this page.</p>
-//     <button href="#" class="sb-header__btn sb-btn sb-btn--primary">Action button</button>
 //   </div>
+//   <button href="#" class="sb-header__btn sb-btn sb-btn--primary">Action button</button>
```
spirosikmd: This looks like a breaking change! Users would need to update their code to move the button outside the header content. Can we avoid that?

dine: We can and the solution is much simpler. The only thing is the title will be aligned a little bit to the top comparing to the back button.

PR #98, opened 2017-09-08
feat(space-zero): Set margin/padding spaces to zero

keepitterron:
Reviewers @dine @piet @spirosikmd @tzengerink

### What
We have different ways to add a margin/padding to an element via a class, but we don't have any way to remove them.
Unfortunately none of the core team members of stylabilla are here to validate it, but I think it's a pretty straightforward concept and safe to add.

### Why
I mostly need it - right now - to remove the gutters from a grid container, which is a much more complex undertaking and I know there are plans already to rework the grid as it's not good enough.

I like the suffix zero as it's immediately meaningful but maybe you have other suggestions?

### Notes for testing
- px2rem has to check if 0 or the linter will complain

### DIFF
--- a/src/objects/_spacing.scss
+++ b/src/objects/_spacing.scss
@@ -10,6 +10,7 @@
 // The available directions are `top`, `bottom`, `left`, `right`, `vertical`, `horizontal`.
 // The available sizes ranges from `xs` to `xl`.
 // The default class `sb-margin` adds the medium sized margin to all sides.
+// The `zero` suffix removes the margin by setting them to 0
 //
 // Markup:
 // <div class="sb-tile sb-margin-bottom-xs sb-padding">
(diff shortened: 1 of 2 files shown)


spirosikmd:
Maybe an idea is to reset the gutters in the project where stylabilla is used? I am not sure if we should add something like this to stylabilla. As you said, the grid is currently under improvements, so this should be fixed?

src/objects/_spacing.scss
```diff
@@ -10,6 +10,7 @@
 // The available directions are `top`, `bottom`, `left`, `right`, `vertical`, `horizontal`.
 // The available sizes ranges from `xs` to `xl`.
 // The default class `sb-margin` adds the medium sized margin to all sides.
+// The `zero` suffix removes the margin by setting them to 0
```
tzengerink: I'm not that familiar with the stylabilla project yet, but maybe renaming it to `-none` makes more sense? That way it's not telling anything about the actual pixels it translates to, `-xs` and all others don't convey the number of pixels either?

keepitterron:
Yeah, I agree this should be address by the grid system. But we didn't plan any time to work on stylabilla and one of the tasks we did depends on either solution.
But this change, while tiny needs to be validated by the team and they're not here.

PR #104, opened 2017-09-19
feat(list): add list component

katia-agna:
list


src/components/_list.scss
```diff
@@ -0,0 +1,25 @@
...
+//    <li> Element 4 </li>
+//  </ul>
+//
+// Styleguide list
+
+.#{$ns}list-table {
+  list-style: none;
+
+  li {
+    border-bottom: 1px solid $sb-color-light-4;
```
spirosikmd: We use exactly this `1px solid $sb-color-light-4` in different places now. Should we create a variable for it?

src/components/_list.scss
```diff
@@ -0,0 +1,25 @@
+// List
+//
+//Markup :
```
endritbajo: Space before the `Markup` and remove the space before the color 😄

src/components/_list.scss
```diff
@@ -0,0 +1,25 @@
...
+//    <li> Element 2 </li>
+//    <li> Element 3 </li>
+//    <li> Element 4 </li>
+//  </ul>
+//
+// Styleguide list
+
+.#{$ns}list-table {
+  list-style: none;
+
```
endritbajo: We need to reset also the default `margin` and `padding` of the list:
```css
  margin: 0;
  padding: 0;
```

src/components/_list.scss
```diff
@@ -0,0 +1,25 @@
+// List
```
spirosikmd: Maybe we should name it Table List?

src/components/_list.scss
```diff
@@ -0,0 +1,25 @@
...
+//    <li> Element 3 </li>
+//    <li> Element 4 </li>
+//  </ul>
+//
+// Styleguide list
+
+.#{$ns}list-table {
+  list-style: none;
+
+  li {
```
spirosikmd: Should we target `li` elements directly or through `table-list__item` class?

src/components/_list.scss
kdabo: I think I would also call it just `list`

src/components/_list.scss
```diff
@@ -0,0 +1,25 @@
+// List
+//
+//Markup :
+//  <ul class="sb-list-table">
```
kdabo: I would maybe have a class `sb-list` here and than for each item in the list, I would add `sb-list__item`

src/components/_list.scss
kdabo: do we also need to add padding for a list item?

src/components/_list.scss
beerecca: We discussed this earlier - we should call it Table List or something else, because at some point we will want to add styling for a normal list component, bullet points etc.

katia-agna: yes I agree with @beerecca and @spirosikmd I will call it table-list

src/components/_list.scss
katia-agna: I'm not a big fan of adding many elements. But on the other hand, we add an element for each `<li>` inside the dropdown element. So I would go for the same logic everywhere

src/components/_list.scss
katia-agna: @kdabo we want to have just the container and the hover action, we don't want to have any restriction like padding that we would have to overwrite if we display different type of information.  (But I agree it looks a little ugly right now)

src/components/_list.scss
katia-agna: aaaaargh !  I'm so bad with that, Thank you :)

src/components/_list.scss
katia-agna: ok yes I agree I'll add an class for each ite,

src/components/_list.scss
beerecca: You can add `sb-padding` to the example in the documentation so it doesn't look ugly :)

src/components/_table-list.scss
```diff
@@ -0,0 +1,27 @@
...
+//
+// Markup:
+//  <ul class="sb-table-list">
+//    <li class="sb-table-list__item"> Element 1 </li>
+//    <li class="sb-table-list__item"> Element 2 </li>
+//    <li class="sb-table-list__item"> Element 3 </li>
+//    <li class="sb-table-list__item"> Element 4 </li>
+//  </ul>
+//
+// Styleguide list
```
spirosikmd: Do we need to specify `table-list` here?

katia-agna: ha yes good catch. I think we need yes

PR #105, opened 2017-09-20
feat(typography): add subtitle element

katia-agna:
screen shot 2017-09-20 at 14 40 40


src/components/_text.scss
```diff
@@ -5,14 +5,15 @@
...
-// .sb-title--s       - Used for smaller titles
-// .sb-text           - Used for buttons and empty state text
-// .sb-text--s        - Used for all input controls
-// .sb-text--emphasis - Used for emphasizing text
+// .sb-display--l      - Used for single dates when visualizing data
+// .sb-display         - Used for big numbers when visualizing data
+// .sb-title--l        - Used for page titles
+// .sb-title           - Used for main titles in different page sections
+// .sb-title--s        - Used for smaller titles
+// .sb-title--subtitle - Used for subtitle
```
spirosikmd: This will not always be a subtitle, right? I remember we mentioned the class `sb-text--subtle`? We use similar text style in campaign overview page. @galagmon?

katia-agna: I am open to any better name than mine

spirosikmd: It is not so much about the name, but more about what it should represent.

endritbajo: The other text classes have a pattern of naming which is: `sb-title-(s,l)`, `sb-text-(s,l)` where `s` and `l` represent sizes. I would vote to name it `sb-subtitle` so we're consistent with the other classes. Even though it shares the same size as the `sb-text` it has a different meaning.

spirosikmd: But then we have `.sb-text--emphasis` which is neither `s` nor `l`?

galagmon: I would maybe think of sb-title-light or sb-title-subtle and then we can use the size if needed. What do you think?

endritbajo: Oh I missread your comment. I read `sb-text--subtitle` instead of `sb-text--subtle` 😬
In that case yes, that could be a valid option.

spirosikmd: I would go for `sb-text--subtle`. The `sb-text` we use in the header is the same as we use in the highlight section.

PR #107, opened 2017-09-21
fix(table-list): last-row-style

katia-agna:
## ChangeLog
- Remove the border bottom for the last child

screen shot 2017-09-21 at 11 05 52


src/components/_table-list.scss
```diff
@@ -23,5 +23,9 @@
     &:focus {
       background-color: $sb-color-light-4;
     }
+
+    &:last-child {
```
endritbajo: Maybe we can avoid overriding by doing:

```css
&:not(:last-child) {
  border-bottom: 1px solid $sb-color-light-4;
}
```

katia-agna: good idea !

PR #111, opened 2017-09-21
feat(upload): long file names overflow gracefully

pietvanzoen:
Closes: https://usabilla.atlassian.net/browse/[MALL-388](https://usabilla.atlassian.net/browse/MALL-388)

Reviewers: @dine @keepitterron @pietvanzoen @beerecca 

dst52

* Tried using `overflow: scroll` but the scroll bar overlaying the filename was quite unsightly.
* Despite the html changing this is backwards compatible with the previous implementation.
* Tested on IE 10+, FF, Safari, Chrome. 
* Only tested with VoiceOver as that's all i have locally.

### DIFF
--- a/src/components/_upload.scss
+++ b/src/components/_upload.scss
@@ -14,7 +14,8 @@
 //   <input class="sb-upload__input" type="file" id="upload-id-{{modifier_class}}" name="upload-1" accept="image/*">
 //   <label class="sb-upload__label" for="upload-id-{{modifier_class}}">
 //     <div class="sb-upload__faux sb-input" role="button" tabindex="0" title="Upload file">
-//       <p class="sb-upload__filename">filename.jpg<a class="sb-upload__link" id="upload-link-{{modifier_class}}">change</a></p>
+//       <input class="sb-upload__filename" readonly value="a-very-very-very-very-very-very-very-very-very-very-very-very-long-filename.jpg"/>
+//       <a class="sb-upload__link" id="upload-link-{{modifier_class}}">change</a>
 //     </div>
 //   </label>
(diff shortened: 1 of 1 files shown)


src/components/_upload.scss
```diff
@@ -14,7 +14,8 @@
 //   <input class="sb-upload__input" type="file" id="upload-id-{{modifier_class}}" name="upload-1" accept="image/*">
 //   <label class="sb-upload__label" for="upload-id-{{modifier_class}}">
 //     <div class="sb-upload__faux sb-input" role="button" tabindex="0" title="Upload file">
-//       <p class="sb-upload__filename">filename.jpg<a class="sb-upload__link" id="upload-link-{{modifier_class}}">change</a></p>
+//       <input class="sb-upload__filename" readonly value="a-very-very-very-very-very-very-very-very-very-very-very-very-long-filename.jpg"/>
```
beerecca: We probably shouldn't have multiple `input`'s in this component?

pietvanzoen: I would love to understand why. I was trying to find negative side effects, but couldn't find anything in testing or find anything online. It's valid HTML and effectively solves the problem.

keepitterron: I personally like to use HTML elements in creative ways when it helps, and in this case it does the trick really well and I don't see any downsides. As per specs it doesn't look like an illegal use of it and it passes validation.

I'm not sure if this could affect accessibility, but from a quick search it doesn't look like so.
For me it's a 👍

beerecca: Ok, if it works and accessibility isn't affected (@katia-agna any issues with this you think?) then all good.

PR #113, opened 2017-09-22
fix(table-list): add border bottom when one element

katia-agna:


src/components/_table-list.scss
```diff
@@ -17,14 +17,15 @@
...
+    border-bottom: 1px solid $sb-color-light-4;
 
     &:hover,
     &:focus {
       background-color: $sb-color-light-4;
     }
 
-    &:not(:last-child) {
-      border-bottom: 1px solid $sb-color-light-4;
+    &:not(:only-child):last-child {
```
spirosikmd: I think we can keep it as before and add another selector for only child, for example

```scss
&:not(:last-child),
&:only-child {
  border-bottom: 1px solid $sb-color-light-4;
}
```

katia-agna: it doesn't work. But I think stylabilla is doing a weird thing.

spirosikmd: I tried locally and it worked for me!

PR #120, opened 2017-10-11
feat(flex): added flex alignment options

endritbajo:
Added some generic and reusable flex utilities for managing the alignment of elements.

## [table-list] Before
before

## [table-list] After
after

## Flex Section
flex page


beerecca:
To be honest I'm not sure - a lot of this functionality is now duplicating features of the grid... I would worry that people might just use these classes for their layout rather than using the grid, meaning we get inconsistent layouts all over again. I do like it though, and can see places where you might not want to use the grid columns... Maybe we just need to include some really good documentation on when you should use these classes.

For example, on the button creator we have these width and height fields, and I used the grid (`sb-col-m-4`) to make them take a third of the space. Instead I could have used `sb-justify-content-start`. Which is the best choice? 

screenshot

beerecca:
@endritbajo @spirosikmd was expecting more documentation on this...?

endritbajo:
@beerecca As it needs some more discussion on the documentation we created a ticket for it. It will be added in a next step.

beerecca:
As part of your current sprint yea? Personally I don't think it's a separate task, and that documentation is part of the definition of done for our components.

PR #124, opened 2017-10-26
feat(colors): expose colors as stand alone classes

spirosikmd:


src/settings/_colors.scss
```diff
@@ -15,6 +17,8 @@ $sb-color-dark-1: #23272a;
 $sb-color-dark-2: #41474c;
 $sb-color-dark-3: #59636b;
 
+@include sb-colors-from-variables(($sb-color-dark-1 $sb-color-dark-2 $sb-color-dark-3), dark);
```
spirosikmd: Any idea how to use these classes as documentation? Should it be a separate section with these classes applied to the same text?

src/tools/_functions.scss
```diff
@@ -11,6 +11,18 @@
   @return ($px / $base-font-size) * 1rem;
 }
 
+// This mixin creates color classes out of the given color-array and color-type
+
+@mixin sb-colors-from-variables($color-array, $color-type) {
```
beerecca: Why can't we just use the value from the color-array directly? Rather than passing the color-type? ie `sb-colors-from-variables(($sb-color-darkred-1 $sb-color-darkred-2 $sb-color-darkred-3))`
This function relies on the function being passed it's variables in the correct order which is pretty fragile. Also means we could do black and white this way which is more consistent.

PR #126, opened 2017-10-27
fix(upload): improve accessibility on the upload component

beerecca:
eslint-plugin-jsx-a11y which I'm adding to react-stylabilla complains that the `a` doesn't have an href property which is legit, so I'm changing it to a `p` tag. Not a breaking change because it will look the same whether you use `a` or update to `p`.


src/components/_upload.scss
```diff
@@ -17,7 +17,7 @@
 //   <label class="sb-upload__label" for="upload-id-{{modifier_class}}">
 //     <div class="sb-upload__faux sb-input" role="button" tabindex="0">
 //       <input class="sb-upload__filename" aria-labelledby="upload-button-{{modifier_class}}" readonly value="a-very-very-very-very-very-very-very-very-very-very-very-very-long-filename.jpg"/>
-//       <a class="sb-upload__link" id="upload-link-{{modifier_class}}">change</a>
+//       <p class="sb-upload__link" id="upload-link-{{modifier_class}}">change</p>
```
spirosikmd: `span` would make more sense to me to tell you the truth.

beerecca: good point, have updated

PR #128, opened 2017-10-30
fix(table-list): add border bottom to all list item

katia-agna:


src/components/_table-list.scss
```diff
@@ -19,16 +19,12 @@
...
     background-color: $sb-color-white;
+    border-bottom: 1px solid $sb-color-light-4;
 
     &:hover,
     &:focus {
       background-color: $sb-color-light-4;
     }
-
-    &:not(:last-child),
-    &:only-child {
```
beerecca: Why do you not want a border on the last item, but you do want it if there's only one item?

katia-agna: We evaluated with Gal that the new feature we want to implement on this component (show more/less behaviour) would be nicer if there were a border-bottom on the last element. 
So it means on the last element and if there is just one element we don't need a special rule, it can apply the normal border-bottom rule.

PR #129, opened 2017-10-31
feat(cursors): added cursor util classes

endritbajo:
cursor


src/trumps/_cursor.scss
```diff
@@ -0,0 +1,28 @@
...
+// .sb-cursor - Default cursor
+// .sb-cursor-pointer - Pointer cursor
+// .sb-cursor-not-allowed - Action not allowed cursor
+//
+// Styleguide cursors
+
+.#{$ns}cursor {
+  cursor: default;
+
+  &-pointer {
```
beerecca: These should be `--pointer` and `--not-allowed` because they're variations on a cursor right?

spirosikmd: I had the same question, but then thinking about it, we won't really use `<a class="cursor cursor--pointer">` right? We would only use `<a class="cursor-pointer">`. So the question is does `<a class="cursor--pointer">` makes more sense?

beerecca: Good point!

PR #130, opened 2017-11-01
docs(styles): avoid loading styles twice

dine:
Kss styles already included in {{{styles}}}


beerecca:
Unfortunately this doesn't appear to fix it. Manage to reproduce the bug - it happens when you run `yarn build` then `yarn start` (no matter if that {{{styles}}} line is there or not).
screen shot 2017-11-02 at 16 24 13

dine:
@beerecca made the wrong choice, keeping {{{styles}}} instead of <style...> seems to work now, even after running `yarn build` then `yarn start`

beerecca:
But that doesn't work if you just run `yarn build` :) This is how the docs are deployed and loaded (not via webpack dev server). Sorry! I think the solution is going to be a bit more tricky and involve the webpack config, which is why I gave it 3 story points.

PR #134, opened 2017-11-09
feat(grid): add responsive offset

katia-agna:
screen shot 2017-11-13 at 10 55 57

screen shot 2017-11-13 at 10 56 13


src/objects/_layout.scss
```diff
@@ -291,6 +291,8 @@ $sb-gutter-map: ('-0': 0px, '-s': 11px);
 // Offset
 //
 // The `sb-offset-{i}` classes can be used to move a column to the right.
+// To create a mobile-first responsive offset you can add apply  one of the responsive rule class: Add `s`,`m`,`l` or `xl` rule to the class
```
katia-agna: Not convince about my not beautiful sentence...

spirosikmd: Maybe rename according to above line?

```
The `sb-offset-{size}-{i}` classes can be used for responsive layouts respectively.
```

src/objects/_layout.scss
```diff
@@ -291,6 +291,8 @@ $sb-gutter-map: ('-0': 0px, '-s': 11px);
...
+// example : `sb-offset-xl-#{i}`
```
spirosikmd: Do we need this example here? We didn't provide one for the other case.

src/objects/_layout.scss
```diff
@@ -315,6 +323,38 @@ $sb-gutter-map: ('-0': 0px, '-s': 11px);
   }
 }
 
+@include breakpoint-s {
+  @for $i from 1 through $sb-grid-columns {
+    .#{$ns}offset-s-#{$i} {
+      margin-left: calc((100% / #{$sb-grid-columns}) * #{$i} + (#{$sb-default-gutter} / 2));
+    }
+  }
```
spirosikmd: This block of code is too repetitive. The only thing that changes is the size, `s`, `m`, `l`, `xl`. Maybe we can create a function in `tools/_functions.scss` something like

```scss
@function sb-offset-size($grid-columns, $default-gutter, $size) {
  @for $i from 1 through $grid-columns {
    .#{$ns}offset-#{$size}-#{$i} {
      margin-left: calc((100% / #{$grid-columns}) * #{$i} + (#{$default-gutter} / 2));
    }
  }
}

// and can be used as
@include sb-offset-size($sb-grid-columns, $sb-default-gutter, 's');
```

What do you think?

katia-agna: ok yes, make sense. So should I refactor also the grid responsive function ?

spirosikmd: Hmm, I am not sure about this one. What do you think @beerecca @endritbajo?

katia-agna: btw it doesn't work. I'm trying to figure why

PR #135, opened 2017-11-09
Update docs navigation

kdabo:
- Hidden the level 3 sub sections and moved them as tabs of the level 2 sections.
- Updated the navigation structure to the following:
```
- Behaviours
	- Hovers
	- Cursor
- Components
	- Button 
	- Button Group
	- Checkbox
	- Dropdown
	- Form Control
	- Header
	- Icons
	- Input
	- Label
	- Loader
	- Preview
	- Select
	- Table List
	- Tile
	- Toast
	- Upload
- Layout
	- Containers
	- Grid
	- Flex
	- Responsive
	- Spacing
- Style
	- Colors
	- Typography
```
## Behaviours
screen shot 2017-11-15 at 11 56 47

## Components
screen shot 2017-11-15 at 11 56 59

## Layout
screen shot 2017-11-15 at 11 57 12

## Styles
screen shot 2017-11-15 at 11 57 27


docs-builder/index.hbs
```diff
@@ -21,33 +21,28 @@
...
+                    <span class="kss-nav__name">{{header}}</span>
                 </a>
-              </li>
-            {{/each}}
-            </ul>
-          {{/if}}
-        {{/if}}
-      </li>
-    {{/each}}
+                {{#if isActive}}
```
kagnamazian: Can't you do  `{{#if isActive && children}}` ?

docs-builder/kss-assets/kss.scss
```diff
@@ -266,17 +274,28 @@ $kss-vertical-rhythm      : $kss-font-size * 1.5;
       display: list-item;
       padding-right: 0;
     }
+
+    .kss-section--depth-2,
+    .kss-section--depth-3 {
+      text-transform: capitalize !important;
```
kagnamazian: oula ! no choice that we had  to put `!important` ?

enbajo: You're right, and also the `kss-section--depth-3` is not needed, they're leftovers from the previous commit.

docs-builder/index.hbs
enbajo: Unfortunately there is not way in handlebars to incorporate logical operators into the standard conditional operator 😕

enbajo:
Yeah I opted for not hacking it as the page was also flashing every time you changed the section. I talked with Gal about keeping it in a single page as it is currently and she was fine with it. I'll put some space and a line at the end of each section.

Unfortunately the kss generates everything in the same page and the only way to change the layout is by hacking it. I think moving forward we should go for more flexible options such as:

- https://github.com/straker/livingcss
- https://github.com/DSSWG/DSS

These will return a JSON representation of the CSS docs which gives us more flexibility on how we want to display the styleguide (we can potentially use React).

> Can we also have the top level headings (behaviour, components etc) have a capital first letter?
You mean the ones in the sidebar menu? The headings in the page already have a capital first letter

beerecca:
👍 

No on the page - this is what I get:
screen shot 2017-11-17 at 12 40 26

enbajo:
@beerecca 

The 3rd level menu shouldn't exist, have you pulled the latest?

beerecca:
Nice!

PR #136, opened 2017-11-13
fix(grid): responsive offset not working without size indication anymore

kagnamazian:


kagnamazian:
how do I check that @beerecca ? I'm checking this file https://static.usabilla.com/stylabilla/v5.1.0/stylabilla.css but doesn't give me a lot of input...

kagnamazian:
@beerecca thanks for the feedback. I changed my code and the output is now the same as the grid responsive.

PR #137, opened 2017-11-20
feat(icon): Add icon--link class to icons

kagnamazian:
arrow_hover

### DIFF
--- a/src/components/_icons.scss
+++ b/src/components/_icons.scss
@@ -3,6 +3,7 @@
 // The Stylabilla icon system is used by placing svg icons inline with the class `sb-icon`. Our approach is based on [these](https://css-tricks.com/pretty-good-svg-icon-system/) [articles](https://css-tricks.com/creating-svg-icon-system-react/).</br>
 // If the icon is for decoration only, please add `aria-hidden="true"` to the icon. Otherwise please add a label - for instance `aria-label="Delete"`.</br>
 // Colors can be applied with `sb-icon--success`, `sb-icon--warning`, `sb-icon--error`.</br>
+// Link styling can be applied with `sb-icon--link`.</br>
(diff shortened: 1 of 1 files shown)


src/components/_icons.scss
```diff
@@ -2,7 +2,7 @@
 //
 // The Stylabilla icon system is used by placing svg icons inline with the class `sb-icon`. Our approach is based on [these](https://css-tricks.com/pretty-good-svg-icon-system/) [articles](https://css-tricks.com/creating-svg-icon-system-react/).</br>
 // If the icon is for decoration only, please add `aria-hidden="true"` to the icon. Otherwise please add a label - for instance `aria-label="Delete"`.</br>
-// Colors can be applied with `sb-icon--success`, `sb-icon--warning`, `sb-icon--error`.</br>
+// Colors can be applied with `sb-icon--success`, `sb-icon--warning`, `sb-icon--error`, `sb-icon--link`.</br>
```
beerecca: Your class doesn't just apply a different colour - I think this needs to be documented on a separate line, something like `Link styling can be applied with sb-icon--link`.

kagnamazian: yes good point !

kagnamazian:
Yes I understand your point of view.  But we want it to behave like a link . The initial state of our icon is still grey/black and will become blue with the action. 
I could make a function which takes the current colour of the icon and make it darker or lighter depending on our design. But it'll look like more like an act of "click" than an indication for the user that it is a "link". I don't know if I'm really clear, what do you think ?

kagnamazian:
@galagmon for consistency, Bex and me think we should follow the code colours of a link hover state. The colour from OneVision doesn't  correspond to any colours. would it work for you?

PR #138, opened 2017-11-27
feat(animations): define generic animations

dine:
Applied already to loader:


src/layout/_animations.scss
```diff
@@ -0,0 +1,145 @@
...
+    transform: rotate(0deg);
+  }
+
+  100% {
+    transform: rotate(360deg);
+  }
+}
+
+@keyframes sb-expand {
+  0% {
```
keepitterron: This animation need some work imho.
It's definitely tricky and I'm not sure I have a better suggestion for this, but it's worth investing some time researching it a bit more.
Some issues I have with it:
- The expanding animation is visually ugly (it reminds me too much of jquery 😛 )
- Animating the height is a performance drain as [it triggers layout](https://developers.google.com/web/fundamentals/performance/rendering/)
- Animating the max-height does not always respect the timing (it will take whatever timing you give it to animate to `300px`, if the dropdown is - say - 200px, 100px will be spent doing nothing)

Do you think we can manage at least some perf improvements [by using trsnforms](https://jdsteinbach.com/css/snippet-animate-display-transform/)?

src/layout/_animations.scss
```diff
@@ -0,0 +1,145 @@
...
+
+// Easing
+//
+// Adding the classes will change the default easing of the animation.<br />
+// For animations not using transition, `animation-easing` instead of `transition-easing` classes
+//
+// Browser support: IE10+, Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+// <div class="sb-animation--hoverDetach {{modifier_class}} sb-transition-duration-xl sb-justify-content-center sb-align-items-center sb-flex" style="width: 80px; border-radius:50%; background-color: #008aab; width: 75px; height: 75px; color: #fff;">
```
beerecca: You've not defined `sb-transition-duration-xl` anywhere

src/layout/_animations.scss
```diff
@@ -0,0 +1,145 @@
+/* stylelint-disable number-no-trailing-zeros */
```
beerecca: For what was this necessary? I can't see anything that would cause this?

src/layout/_animations.scss
```diff
@@ -0,0 +1,145 @@
...
+// Animations with * use transition.
+//
+// Browser support: IE10+, Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+// <div class="sb-padding {{modifier_class}} sb-justify-content-center sb-align-items-center sb-flex" style="width: 80px; border-radius:50%; background-color: #008aab; width: 75px; height: 75px; color: #fff;">
+//   Weeee
+// </div>
+//
+// sb-animation--hoverLift - lift * (feedback button hover)
```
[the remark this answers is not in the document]
beerecca: And what does `feedback button hover` refer to?

src/layout/_animations.scss
dine: Not needed, I added because I was playing copy paste with multiple examples without 0 out there 😊 🙈

src/layout/_animations.scss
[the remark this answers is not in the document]
dine: We need to improve the descriptions, Gal just sent me some better text she also worked on.

src/layout/_animations.scss
```diff
@@ -0,0 +1,145 @@
...
+//
+// Styleguide Layout.animation.animations
+
+$sb-easing-function: cubic-bezier(0.38, 1.59, 0.82, 0.96);
+$sb-transition: sb-get-timing('-s') $sb-easing-function;
+$sb-timing-map: ('-s': 0.2s, '-m': 0.5s, '-l': 0.8s);
+$sb-easing-map: ('-cubic': $sb-easing-function, '-linear': linear, '-ease-in': ease-in, '-ease-in-out': ease-in-out);
+
+@function sb-get-timing($size) {
+  @each $suffix, $timing in $sb-timing-map {
```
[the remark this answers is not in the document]
dine: Declaring it one by one? 🙃
But this is also something we don't need or use yet, just inspired by https://marvelapp.com/styleguide/design/animation, so I guess it could be removed from now and we just add the recommended timing to each interaction. What do you think?

src/layout/_animations.scss
dine: @keepitterron hahaha, I agree the `300px` was a mistake, I think I copied from dropdwon original size and it stayed there. About the uglyness, you can discuss with @galagmon, but I think the idea is to show that the options emerge from the user click on the select dropdown, so the example on the page you send I think we are better without (and it is also waaaay to jquery-ish 😛).

src/layout/_animations.scss
```diff
@@ -0,0 +1,145 @@
...
+// </div>
+//
+// sb-animation--hoverLift - lift * (feedback button hover)
+// sb-animation--hoverDetach - lift and detach * (feedback choice hover)
+// sb-animation--spin - continuous rotation (loader)
+// sb-animation--expand - expand (dropdowns, read more)
+//
+// Styleguide Layout.animation.animations
+
+$sb-easing-function: cubic-bezier(0.38, 1.59, 0.82, 0.96);
```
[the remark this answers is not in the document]
dine: Here: https://github.com/usabilla/stylabilla/pull/138/files#diff-08df69f6780b8052e5e1dddc63a56e32L1

dine:
@beerecca @keepitterron updated PR

beerecca:
Yea same. We can always come back to this in the future if we need it. Nice!

PR #139, opened 2017-11-29
docs(icons): added smiley icons

enbajo:
## Before
screen shot 2017-11-29 at 09 48 38

## After
screen shot 2017-11-29 at 09 47 08


beerecca:
Yes I reckon... but can we not use the new ones? Or is that a case of everything becoming inconsistent in the app?

spirosikmd:
I guess everything will become inconsistent... Should we use these for now, and create a plan to replace all current ones in the app with `Icon` component. Then we can migrate to new ones? Does this make sense?

beerecca:
Sounds great. I know Gal will be happy :)

PR #141, opened 2017-12-06
feat(radio): radio and radio group component

dine:
Add radio and radio group components (aligned with checkbox styles).

Update:
Validation and error message examples added:


src/components/_checkbox.scss
```diff
@@ -5,20 +5,69 @@
...
+  border-color: $sb-color-light-4;
+  box-shadow: none;
+}
+
+%sb-faux-disabled-checked {
+  background-color: $sb-color-light-3;
+  border-color: $sb-color-light-3;
+  border-bottom: 1px solid $sb-color-light-1;
+}
+
```
keepitterron: shouldn't all these placeholder selectors be moved to a shared file? i'm not sure where, but since it's also used by the radio it's weird that it lives here.

src/components/_radio.scss
```diff
@@ -0,0 +1,189 @@
+// Radio
+//
+// A basic radio component
+//
+// Browser support: IE9+ (no transition), Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+// <div>
```
keepitterron: aria attributes are missing/do not apply to these components?

dine: For the simple radios no, because we are using the native elements (so it's not like the button group).

src/components/_radio.scss
```diff
@@ -0,0 +1,189 @@
...
+}
+
+// Group
+//
+// A basic radio group. Apply a `.sb-radio-group` class to the radio wrapper.
+//
+// Browser support: IE9+, Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+// <div class="sb-radio-group">
```
dine: But I guess here we might need the `role="radiogroup"` or a `fieldset` to group them. Also I'll update here with a label and error for the whole group.

dine:
PR updated

src/generic/_shared.scss
```diff
@@ -0,0 +1,64 @@
...
+}
+
+%sb-faux-focus {
+  @include sb-focus-no-border();
+}
+
+%sb-faux-input {
+  position: absolute;
+  left: -999em;
+}
```
keepitterron: i'm not a fan of `999` :D
can we use a different approach here? eg: https://github.com/usabilla/themes-publisher/blob/master/lib/compilers/css-compiler/2/common/_mixins.scss#L144

src/tools/_radios-and-checkboxes.scss
```diff
@@ -0,0 +1,69 @@
+/* stylelint-disable plugin/docs-required */
+
+// Shared styles between radio and checkbox components
+// Obs: please do not reuse classes declared on this file for other components than radio or checkbox.
```
beerecca: @dine this should be `extends` not `classes`

PR #142, opened 2017-12-12
chore(add-src): add src folder to npm so we can import specific files

beerecca:
In the feedback form app we want to be explicit about the stylabilla components we use so that we can keep the file size as small as possible.


beerecca:
@spirosikmd as long as you have the right loaders set up, eg -
```
test: /\.scss$/,
exclude: {
  test: /node_modules/,
  exclude: /stylabilla/,
},
use: ['style-loader', 'css-loader', 'sass-loader'],
```
then you can import whatever file you want like `@import "~@usabilla/stylabilla/src/settings/_colors.scss";`

PR #144, opened 2017-12-15
feat(tags): setup tags component

pietvanzoen:
Part of: [MALL-469](https://usabilla.atlassian.net/browse/MALL-469)

Reviewers: @dine @keepitterron @pietvanzoen @beerecca 

TODO:
- [x] transitions?
- [x] accessibility testing
- [x] Browser testing


src/components/_tags.scss
```diff
@@ -0,0 +1,59 @@
...
+  &__label {
+    background-color: $sb-color-white;
+    border: 1px solid $sb-color-light-3;
+    border-radius: $sb-radius-infinite;
+    color: $sb-color-dark-3;
+    cursor: pointer;
+    display: inline-block;
+    font-size: sb-px2rems(12px);
+    margin: 3px 3px 3px 0;
+    padding: 5px 15px;
```
beerecca: Ideally we should be using the same numbers as in _spacing.scss, to keep with the modular scale. You can use #{map-get($sb-spacing-map, '-s')}. Definitely for padding, though I guess margin will have to stay 3px. But we should be using sb-px2rems here too.

src/components/_tags.scss
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  &__label {
+    background-color: $sb-color-white;
+    border: 1px solid $sb-color-light-3;
+    border-radius: $sb-radius-infinite;
+    color: $sb-color-dark-3;
+    cursor: pointer;
+    display: inline-block;
+    font-size: sb-px2rems(12px);
```
beerecca: Need to check this font size - nowhere else do we have 12px font size. The closest is 13px on the sb-text--s class.

src/settings/_colors.scss
```diff
@@ -127,3 +127,12 @@ $sb-color-red-4: #fcaab2;
...
 $sb-color-darkred-3: #972936;
+
+// Sentiment colors
+//
+// $sb-color-positive: #29c682;
+// $sb-color-negative: #f16371;
+//
+// Styleguide Styla.colors.sentiment
+$sb-color-positive: #29c682;
+$sb-color-negative: #f16371;
```
beerecca: We already have colours in stylabilla that are verrry similar $sb-color-green-2 and $sb-color-red-2. Would much prefer to use these.

beerecca: There's also a typo here - `Styla`

src/components/_tags.scss
beerecca: I've created a function that gets the right spacing in rems, will make a pr with it this morning.

src/components/_tags.scss
```diff
@@ -0,0 +1,67 @@
...
+  &--#{$name} &__input:checked + &__label {
+    background-color: $state-color;
+    border-color: $state-color;
+    color: $sb-color-white;
+  }
+}
+
+.#{$ns}tags {
+  &__input {
+    @extend %sb-faux-input;
```
dine: If we are going to use [`%sb-faux-input`](https://github.com/usabilla/stylabilla/blob/master/src/tools/_radios-and-checkboxes.scss#L32) for tags too I think it would be better to move it to something more generic than the file `_radios-and-checkboxes.scss`.

src/components/_tags.scss
```diff
@@ -0,0 +1,67 @@
...
+    cursor: pointer;
+    display: inline-block;
+    font-size: sb-px2rems(13px);
+    margin: sb-px2rems(3px);
+    margin-left: 0;
+    padding: sb-px2rems(map-get($sb-spacing-map, '-xs')) sb-px2rems(map-get($sb-spacing-map, ''));
+  }
+
+  &__input:focus + &__label {
+    @extend %sb-faux-focus;
```
dine: Same here.

src/components/_tags.scss
```diff
@@ -0,0 +1,67 @@
...
+    background-color: $sb-color-white;
+    border: 1px solid $sb-color-light-3;
+    border-radius: $sb-radius-infinite;
+    color: $sb-color-dark-3;
+    cursor: pointer;
+    display: inline-block;
+    font-size: sb-px2rems(13px);
+    margin: sb-px2rems(3px);
+    margin-left: 0;
+    padding: sb-px2rems(map-get($sb-spacing-map, '-xs')) sb-px2rems(map-get($sb-spacing-map, ''));
```
dine: Replace by https://github.com/usabilla/stylabilla/pull/149/files#diff-8ef85da964e2daaaaec6e723197f036aR16

src/components/_tags.scss
beerecca: @pietvanzoen @dine @keepitterron We have documentation in the radios-and-checkboxes file that specifically says the %sb-faux-input and %sb-faux-focus should not be used for components other than radios and checkboxes. I can see faux-input has also been used on the rating component. If there's a need to use them elsewhere then they should be moved from the radios-and-checkboxes file, or the file should be renamed and documentation updated. 

I propose we:
a) Change %sb-faux-input into a mixin, and move to a new file - tools/mixins
b) Move all mixins from tools/functions, and the two focus mixins in settings/box, to the new tools/mixins (all other mixins are pretty component specific so should stay where they are)
c) Use sb-focus-no-border() instead of %sb-faux-focus in ratings and tags components
And leave the radios and checkboxes file as it is, an exception because these components are so similar. Maybe adding a linter to prevent these from being used again :)

Reasoning being, the more we use mixins/extends etc, especially an extend that extends a mixin, the more likely we are to get crappy compiled css, the harder it is to quickly reason about, the barrier to entry gets higher... https://csswizardry.com/2014/11/when-to-use-extend-when-to-use-a-mixin/

And in fact, I've just checked the compiled css and currently all the extends are being compiled at the top of the file, before even the reset - not ideal. So I think we should also change the extends in the radios and checkboxes file to mixins, as per Harry's recommendation. And maybe the animations too, then we can add a stylelint for "at-rule-blacklist" to blacklist @extend.

What do you think?

dine: I thought about it and I even commented on this PR, but then behind the tag we were also using a checkbox so in a way it made sense. The extends being at the top at the file is something that I didn't notice, though. Because of that I think we can go with the blacklist @extend and replace everything to mixins in this case. It might generate more lines in some places (like the radios/checkbox example), but I think it will be safer.

PR #145, opened 2017-12-15
refactor(imports): group imports for easier selective imports

beerecca:

Grouping the imports this way allows for us to more easily selectively import components that rely on most of the settings, tools, elements etc - otherwise we'd have to list every one of those files out. They can still of course be imported separately.


pietvanzoen:
Am i correct in thinking that if someone wanted to import one of the groups they'd also have to import 'namespace'?

beerecca:
@pietvanzoen yes

PR #147, opened 2017-12-19
feat(validation): Apply general form control validation

keepitterron:
General style validation for the form control. `react-stylabilla` is already good to go, we just have to update the stylabilla version once this is published.

- `--invalid` applies a red border and a red validation message
- `--valid` has been removed
- radio, rating/nps, checkboxes only get a red error message

**THIS IS THE UPDATED VERSION**
screen shot 2017-12-19 at 17 39 40

**wrong version of red and wrong font, updated in this PR, still here for reference.**
input/textarea
screen shot 2017-12-19 at 15 45 31

upload
screen shot 2017-12-19 at 15 46 03

select/dropdown
screen shot 2017-12-19 at 15 45 46

Close [MALL-499](https://usabilla.atlassian.net/browse/MALL-499)
Reviewers @usabilla/mallard @galagmon

### DIFF
--- a/src/components/_form-control.scss
+++ b/src/components/_form-control.scss
@@ -23,13 +23,14 @@
   &__validation {
     display: none;
     margin-top: sb-px2rems(7px);
+    font-family: $sb-base-font-family;
     font-size: sb-px2rems(13px);
-    color: $sb-color-red-1;
+    color: $sb-color-red-2;
   }
 
   &--invalid {
     .#{$ns}input {
-      border: 1px solid $sb-color-red-1;
+      border: 1px solid $sb-color-red-2;
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/components/_form-control.scss
```diff
@@ -11,70 +11,31 @@
...
-// .sb-form-control--success - success input
 //
 // Styleguide Components.form
 
 .#{$ns}form-control {
   &__validation {
     display: none;
     margin-top: sb-px2rems(7px);
+    font-family: $sb-base-font-family;
+    font-weight: 400;
```
spirosikmd: Do we need to specify `font-weight` here? Isn't this the default?

keepitterron:
I think we don't need a breaking change. Nothing needs to be changed, it's just visual styles.
And as far as I know we do not rely on the previous validation in any ways, it won't be confusing for users I guess. Am I missing something?

spirosikmd:
I guess it is a breaking change. If someone has implemented the previous structure with the icon, then they assume the styles that target this icon are still there. Upgrading would break their icon styles. They would need to update their Vue component library, for example, to not use an icon anymore, right?

spirosikmd:
For breaking change and semantic release to work, we should use it exactly like `BREAKING CHANGE:` (notice the colon at the end).

keepitterron:
doesn't say so in the CONTRIBUTING docs, we should update it if the colon is needed.

spirosikmd:
True! Fixed in #148 @keepitterron

PR #148, opened 2017-12-21
docs(contributing): update breaking change contribution

spirosikmd:


spirosikmd:
This is how [angularjs](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#commit-message-format) specifies this

```
<type>(<scope>): <subject>
<BLANK LINE>
<body>
<BLANK LINE>
<footer>
```

and the footer we should add any breaking change related information. In terms of semantic release it doesn't matter where you place the `BREAKING CHANGE:` so I guess we can leave as is?

keepitterron:
After reading the documentation for semantic release, my comment about the colon is just stupid :D there's no need to add the colon in this docs.
There are examples provided to show exactly [how to structure your commit message](https://github.com/semantic-release/semantic-release#default-commit-message-format). it's super clear, I was just lazy.

Maybe we can link to this just to be extra explicit, otherwise you can just close this PR, I just had the assumption that the keyword `BREAKING CHANGE` was only used to bump the major and nothing else.

PR #149, opened 2017-12-21
refactor(settings): split settings from css output

beerecca:
BREAKING CHANGE: Pull out css output from settings/index and tools/index so that when the feedback form app imports these settings, css output won't be duplicated.

Fix close icon focus.

Add function to get spacing in rems based on size keyword.

Add support for rtl.


src/tools/_functions.scss
```diff
@@ -6,15 +6,21 @@
...
   @return ($px / $base-font-size) * 1rem;
 }
 
-// This mixin creates color classes out of the given color map
+// This function returns a rems value based on the given size keyword (eg: '-xs', '-s')
+@function sb-rem-spacing($size) {
+  $base-font-size: 16px;
+  $px: map-get($sb-spacing-map, '-#{$size}');
 
+  @return ($px / $base-font-size) * 1rem;
```
keepitterron: can't we just use `sb-px2rems` after calculating the `$px` mappings?

dine: done 👍

PR #151, opened 2018-01-04
feat(component): tooltip component

keepitterron:
screen shot 2018-01-04 at 16 43 32
A minimal tooltip used for enhancing interactions.

Shitty animated gif
tooltip

Closes [MALL-439](https://usabilla.atlassian.net/browse/MALL-439)
Reviewers @usabilla/mallard @galagmon


src/components/tooltip.scss
```diff
@@ -0,0 +1,126 @@
...
+  border-radius: 2px;
+  box-shadow: $sb-box-shadow-light;
+  font-family: $sb-base-font-family;
+  font-size: sb-px2rems(13px);
+  line-height: sb-px2rems(19px);
+  color: $sb-color-dark-3;
+  max-width: sb-px2rems(350px);
+  transition: opacity $sb-timing-s ease-out, transform $sb-timing-m $sb-easeup-function;
+  opacity: 0;
+  pointer-events: none;
```
dine: Could this be a problem for example if we have links on the privacy text?

src/components/tooltip.scss
```diff
@@ -0,0 +1,126 @@
...
+// Tooltip element default (medium).
+// Used for descriptions or explanation of an action on the interface.
+//
+// Browser support: IE9+, Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+// <div class="sb-tooltip__container">
+// <div>hover me</div>
+// <div class="sb-tooltip">
+//   <div class="sb-tooltip__content">
```
dine: There are no styles applied on `sb-tooltip__content`, we can remove it.

galagmon:
Its hard for me to see through the gif, I think it looks ok. Please call me tomorrow to see

keepitterron:
This is updated.
I didn't add the modifier class 'cause its not needed. there are two separate examples for small and large because also the content changes.

keepitterron:
screen shot 2018-01-08 at 16 02 11
screen shot 2018-01-08 at 16 02 01
screen shot 2018-01-08 at 17 10 13

keepitterron:
Issues that took me too much time already:

since the tooltip is positioned absolutely into a relative container, if the container has a fixed width, the tooltip will not grow outside the container if we don't specify a fixed width.
Solution: small tooltips will always stay on one line and extend automagically, big/medium tooltips will find a solution when we implement it. maybe adding a custom class with the sizing.

Fucking firefox needs a `1px` adjustment for the arrow.

To be tested on IE 9/10, yet.

src/components/_tooltip.scss
```diff
@@ -0,0 +1,221 @@
+// Tooltip
+//
+// Tooltip element default (medium).
+// Used for descriptions or explanation of an action on the interface.
+//
+// Browser support: IE9+, Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+// <div class="sb-tooltip__container">
+// <div class="hover-area sb-text--center">hover me</div>
```
beerecca: Can remove `hover-area`? And these two divs could be a single div with `sb-tile sb-tooltip__container sb-text--center` on it for the documentation?

src/components/_tooltip.scss
```diff
@@ -0,0 +1,221 @@
...
+// <div class="sb-tooltip {{modifier_class}}">
+//   <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vestibulum vestibulum turpis enim, eget tempor tellus placerat at.</p>
+//   <p>Alea iacta est!</p>
+//   <div class="sb-tooltip__arrow"></div>
+// </div>
+// </div>
+//
+// .sb-tooltip--bottom - Position it on the bottom
```
beerecca: Can we put `Position the arrow on the bottom`?

src/components/_tooltip.scss
```diff
@@ -0,0 +1,221 @@
...
+//
+// Styleguide Components.tooltip
+
+$tp: #{$ns}tooltip;
```
beerecca: Should use this consistently everywhere in this component

src/components/_tooltip.scss
```diff
@@ -0,0 +1,221 @@
...
+
+.#{$ns}tooltip {
+  position: absolute;
+  display: table;
+  top: 0;
+  left: 50%;
+  transform: translate(-50%, -100%);
+  padding: sb-px2rems(11px);
```
beerecca: Can use `sb-rem-spacing('-s')`
Also most other places you're including padding.

src/components/_tooltip.scss
```diff
@@ -0,0 +1,221 @@
...
+  background: $sb-color-white;
+  border: 1px solid $sb-color-light-3;
+  border-radius: 2px;
```
beerecca: Can use `$sb-radius`

src/components/_tooltip.scss
```diff
@@ -0,0 +1,221 @@
...
+  box-shadow: $sb-box-shadow-light;
+  font-family: $sb-base-font-family;
+  font-size: sb-px2rems(13px);
```
beerecca: Can use @dine's new mixin

src/components/_tooltip.scss
```diff
@@ -0,0 +1,221 @@
...
+  opacity: 0;
+  pointer-events: none;
+
+  &--bottom {
+    top: 100%;
+    left: 50%;
+    transform: translateX(-50%);
+  }
+
+  &--left {
```
beerecca: This is a little confusing to have the classes in here and all the documentation at the bottom - if it's not a simple variation on the component (ie `sb-preview--active`) then the classes should be included with the separately documented component (ie `Components.button.loader`).

src/components/_tooltip.scss
```diff
@@ -0,0 +1,221 @@
...
+    &:hover .#{$tp}--left {
+      transform: translate(0, calc(-100% - 7px));
+    }
+
+    &:hover .#{$tp}--right {
+      transform: translate(0, calc(-100% - 7px));
+    }
+  }
+
+  & > * {
```
beerecca: I don't think we should include this, it forces it so strongly. Personally I think the font size on the sb-tooltip is enough.

src/components/_tooltip.scss
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    .#{$tp}--bottom &::after {
+      transform: translate(-50%, 7px) rotate(45deg);
+    }
+  }
+}
+
+.#{$tp}__arrow {
+  @include sb-firefox-only {
```
beerecca: Gross but I think the implementation is good :)

src/components/_tooltip.scss
```diff
@@ -0,0 +1,221 @@
...
+
+    .#{$tp}--bottom & {
+      top: 1px;
+    }
+  }
+}
+
+// Tooltip left
+//
+// Tooltip element small. Always displayed on one line!
```
beerecca: Typo - left or small? Same with right

src/components/_tooltip.scss
```diff
@@ -0,0 +1,221 @@
...
+// Tooltip right
+//
+// Tooltip element small. Always displayed on one line!
+// Used for small descriptions or names. eg: emotions in mood rating.
+//
+// Browser support: IE9+, Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+// <div class="sb-tooltip__container">
+// <div style="text-align: right">hover me</div>
```
beerecca: Really prefer not to include this even though it's for the documentation, it's so much easier for someone to accidentally copy. Could add a class to trumps/text-align

src/components/_tooltip.scss
keepitterron: not in master yet, good point to remember

src/components/_tooltip.scss
keepitterron: I'd say size and positioning are indeed small variations.
The reason why they have separate documentation at the bottom is the tool we use for documentation that is limiting. They're very well self-descriptive.
You disagree?

src/components/_tooltip.scss
keepitterron: In theory you could put whatever markup inside the tooltip, and the sizing should always be what the design indicates.
All `<p>`s in stylabilla have a font-size defined with a very high specificity `p { font-size: 1rem }` I need to override that with this ugly bit. while i'm at it i've made it generic.

src/components/_tooltip.scss
keepitterron: Done that.
But I'm afraid this is even more confusing. Should I include also `sb-tile` and `sb-text--center` in a tooltip implementation? Are those cosmetic classes? Are those there for a reason?
It's not clear, we need a better way to indicate cosmetic classes for documentation's puproses only.
Maybe a `kss` prefix (or suffix)?

keepitterron:
@beerecca updated and replied to some other comments.

src/components/_tooltip.scss
beerecca: Maybe, but for me I think this is pretty clear, maybe it's just cos it's how I'm used to doing it. A component should always be usable on it's own - if say you needed padding in the tooltip then it should be included in the tooltip styles rather than with a sb-padding (which is what we've done) therefore for me it's pretty clear that any additional styles not prefixed by sb-tooltip are there just to illustrate how to combine the component with other things, normally for the sake of documentation. We do a similar thing for the grid etc.

src/components/_tooltip.scss
beerecca: By `simple` I meant that they didn't require additional documentation, for me it just feels weird reading the documentation which just includes "bottom" and then seeing "left" and "right" in the accompanied styles. But I'm not that fussed, you're right, it is well documented in the end.

src/components/_tooltip.scss
beerecca: Ah ok didn't realise sb-tooltip wasn't overriding the p. Hm not ideal that we're already having specificity issues... I guess I'm thinking of customisations that want different sizes in the tooltip... But yea, not that big a deal. Happy to leave for now.

beerecca:
Nice! Yea the couple of comments you've added I'm happy with, looks good to me!

beerecca:
Nice! Looks good to me, I can't tell the difference :)

PR #153, opened 2018-01-05
fix(rating): align with design files

dine:
### Changelog
- Remove extra space when using `sb-rating-group--after`
- Update examples to align with sketch files
- Set proper font and color to high/low labels
- **Bonus:** add and apply `base-font()` mixin

### Before
screen shot 2018-01-05 at 16 53 44

### After
screen shot 2018-01-05 at 16 54 05


src/components/_dropdown.scss
```diff
@@ -27,9 +27,9 @@
 // Styleguide Components.dropdown
 
 .#{$ns}dropdown {
+  @include base-font();
```
beerecca: @dine Shouldn't call this `base-font` because we already have a variable called $base-font-size and it's set to 16px. 

Based on my comments here https://github.com/usabilla/stylabilla/pull/144#discussion_r160176377
I would suggest we call this `sb-text-s` or something and include in the tools/mixins file.

PR #154, opened 2018-01-09
fix(rating): low / high text optional

dine:
Reviewers: @beerecca @keepitterron 

### Changelog
- Added `sb-rating-text` group to docs
- Replace `--after` to `--inline`
- Fix `ratingroup` to `radiogroup`

screen shot 2018-01-11 at 17 28 31
screen shot 2018-01-11 at 17 28 46


src/components/_rating.scss
```diff
@@ -155,13 +155,25 @@ $sb-rating-before-space: sb-px2rems(7px);
 
     position: absolute;
     bottom: 0;
+    display: none;
```
beerecca: I don't think we want this override. The default should just not include the text in the markup - a conditional render in react-stylabilla.
This also means we can remove the display block from below - as per the contributing guidelines "Avoid applying styles in a generic class and then undoing them in a more specific class."

dine: Updated, but the documentation will be a little weird with the text overlapping the rating. I added a note.
screen shot 2018-01-10 at 15 07 36

beerecca: Maybe it would be better to split it in the documentation? Rating with text, rating without text?

dine: I can add another docs without text, but this block we would still have the default styling (with overlapping text), because of this rules. 

They do make sense for most of our components though.

dine:
Ok, PR and screenshots updated

beerecca:
Nice! Sorry about all the back and forth!

PR #156, opened 2018-01-09
refactor(mixins): reorder and replace extends by mixins

dine:
Reviewers: @beerecca @keepitterron @kagnamazian 

Changelog:
- Replace previous `@extends` with placeholder selectors (%) by `@mixins`
- Add `no-extends` rule and lint plugin
- Rename `base-font()` mixin to `sb-text-default()`
- Ignore `no-extends` rule in components using animation

Pending:
- Decision: change something about animation `@extends`?

Based on the comments:
- https://github.com/usabilla/stylabilla/pull/153#discussion_r160198859
- https://github.com/usabilla/stylabilla/pull/144#discussion_r160176377


src/tools/_index.scss
```diff
@@ -1,2 +1,3 @@
 @import 'functions';
-@import 'radios-and-checkboxes';
+@import 'box';
```
keepitterron: I absolutely agree with the need of separating this stuff as much as possible but it's really hard to find where these mixins live. `box` doesn't mean anything to me... wouldn't be better to have a `functions` and a `mixins`?

src/tools/_text.scss
```diff
@@ -0,0 +1,5 @@
+
+@mixin sb-text-s() {
```
keepitterron: if I see a `sb-text-s` I expect a `sb-text-m` and a `sb-text-l`...
wouldn't make more sense to have this be `sb-text-default` or something like that specifying this is the default/base font?

dine:
PR (and description) updated

lint/no-extends.js
```diff
@@ -0,0 +1,25 @@
+const stylelint = require("stylelint");
```
beerecca: We don't need to create our own plugin here, one already exists... I mentioned it in my comment -"at-rule-blacklist"
https://stylelint.io/user-guide/rules/at-rule-blacklist/

src/components/_form-control.scss
```diff
@@ -21,7 +21,7 @@
 
 .#{$ns}form-control {
   &__validation {
-    @include base-font();
+    @include sb-text-default();
```
beerecca: Sorry should've commented on @keepitterron's comment - I disagree with "text-default" because the default text should be paragraph text which is 16px. I much prefer your "text-s" version - and I'm fine with adding a "sb-text" and "sb-text-l" or whatever.

dine:
Ok, updated PR with `sb-text-s` since I have no strong opinions about it. Agree that default could be expected the paragraph font, but if `sb-text-s` is a problem it could also be called `sb-text-alternative` or `sb-text-variant`. I just don't think we should add a "sb-text" and "sb-text-l" before we actually need it.

beerecca:
Nice, yup happy with that :)

PR #157, opened 2018-01-18
feat(mood-rating): add mood rating component

dine:
BREAKING CHANGE: 
- default smiley size is now `30px` instead of `28px`.
- tooltips changed position

## Important
- `--default` variation was created only for purposes of avoiding overrides when we use `--progressive`
- `--default` and `--custom` smilies need to have `--selected` class applied after an option is selected, so it can have different opacity to previous siblings;
- Progressive hover effect works with CSS only, but requires all inputs and labels to be on the same level (no wrappers can be added around `input`+`label`); 

### Braindumping
- `--custom` sounds too generic, since we could have custom stars (or any other image) too;
- In the future, should we have different base markups for progressive hover / non progressive hover with stars in SVG?
- With SVG stars we could use `fill` property to apply the `moodColor` we have on our themes db.

## Screenshots

Icons before/after:
screen shot 2018-01-22 at 17 13 50
screen shot 2018-01-26 at 10 23 19


src/tools/_shared.scss
```diff
@@ -28,7 +28,7 @@
 // hide input
 @mixin sb-faux-input() {
   appearance: none;
-  vertical-align: middle;
+  left: -100%;
```
dine: This was added for Firefox older versions, and still readable by screen readers.

src/components/_mood-rating.scss
```diff
@@ -0,0 +1,268 @@
...
+    @include sb-faux-input();
+  }
+
+  &__label {
+    border-radius: 50%;
+    cursor: pointer;
+    display: inline-block;
+    width: sb-px2rems($sb-mood-size);
+    height: sb-px2rems($sb-mood-size);
+    margin-right: sb-px2rems(23px);
```
keepitterron: we can use `sb-rem-spacing` here and everywhere else it applies

src/components/_mood-rating.scss
```diff
@@ -0,0 +1,268 @@
...
+    .#{$ns}mood-rating__input:checked ~ .#{$ns}mood-rating__label:hover,
+    .#{$ns}mood-rating__input:checked + .#{$ns}mood-rating__label,
+    .#{$ns}mood-rating__input:focus + .#{$ns}mood-rating__label {
+      @extend .#{$ns}animation--hover-lift;
+
+      opacity: $sb-mood-opacity-selected;
+    }
+
+    .#{$ns}tooltip__container:hover .#{$ns}tooltip {
+      transform: translate(-50%, calc(-100% - 14px));
```
keepitterron: this shouldn't be here but in the tooltip component. if we need more spacing or the tooltip does not behave as it's designed we should fix it or add a modifier :)

src/components/_mood-rating.scss
```diff
@@ -0,0 +1,268 @@
...
+
+  &--progressive {
+    .#{$ns}mood-rating__label {
+      background-size: $sb-mood-size * 2;
+      margin-right: sb-px2rems(16px);
+    }
+
+    .#{$ns}mood-rating__input:checked ~ .#{$ns}mood-rating__label,
+    .#{$ns}mood-rating__label:hover ~ .#{$ns}mood-rating__label {
+      background-position: 100% -1px;
```
keepitterron: this feels very weird. I'm sure there's a reason for it but it's very hard to tell.
I'd first try to find a way to not have that `-1px`, if it's not possible and it's absolutely needed I'd add a comment as to why it's there.

src/tools/_shared.scss
keepitterron: I'm not a fan of this. Hiding elements by removing them from the document flow and moving them outside of the viewport has some nasty downsides (eg: making the viewport much bigger, unexpected consequences when focusing an element that is outside the viewport, etc.)
I'd follow the `visually-hidden` concept you introduced below. Or - worst case scenario - just target the shit browser that needs this.

src/components/_mood-rating.scss
```diff
@@ -0,0 +1,268 @@
...
+// - Replace SVG for a `sb-visually-hidden` span with the label.
+//
+// Browser support: IE9+, Safari6.2+, Firefox30+, Chrome32+
+//
+// Markup:
+// <div class="sb-mood-rating sb-mood-rating--custom {{modifier_class}}" role="ratinggroup">
+//   <input type="radio" value="custom-hate" name="mood-rating-custom-{{modifier_class}}" id="mood-rating-hate-custom-{{modifier_class}}" class="sb-mood-rating__input">
+//   <label class="sb-mood-rating__label sb-tooltip__container kss-mood__custom" for="mood-rating-hate-custom-{{modifier_class}}">
+//     <span class="sb-visually-hidden">Hate</span>
+//     <div class="sb-tooltip sb-tooltip--s">
```
keepitterron: keep in mind that the tooltip will wrap an element. if we wrap the label then you can't use the proximity (or whatever `~` is called. if we wrap each individual mood element the markup will be different.

src/components/_mood-rating.scss
dine: It feels weird indeed, but it was mostly because of the focus ring. I'll update with some other things @galagmon asked and it should not be needed anymore.

src/components/_mood-rating.scss
dine: I don't think we need to change the tooltip siblings, we just need the regular `tooltip__container` behaviour now.

dine:
@kagnamazian I think product/design would be happy to eventually update all icons at once, but that is not going to happen (yet). Maybe after we implement Stylabilla everywhere. I checked quickly with @spirosikmd and apparently this icons are not being used yet on the dashboard, or any other place other than react-stylabilla. If we search for all references of `sb-icon--m` [they only exist there](https://github.com/search?utf8=%E2%9C%93&q=org%3Ausabilla+sb-icon--m&type=Code). Also searching for `icon="smiley-5"` for example [also only there](https://github.com/search?utf8=%E2%9C%93&q=org%3Ausabilla+icon%3D%22smiley-5%22&type=Code). Am I missing something?

src/components/_mood-rating.scss
```diff
@@ -0,0 +1,210 @@
...
+    .#{$mr}__icon {
+      visibility: hidden;
+    }
+  }
+
+  &--custom {
+    $sb-mood-full-size: $sb-mood-size * 5;
+
+    .#{$mr}__label {
+      background-image: url('https://d6tizftlrpuof.cloudfront.net/themes/production/test-test-email-campaign-production-mood-custom-sprite-url-32e670f826d1a5460d92714f7604f0d5.png');
```
beerecca: Ideally these would be in the assets folder and therefore uploaded to s3 alongside these files, meaning we can have versioned images. Check how the fonts are referred to for an example.

src/components/_mood-rating.scss
```diff
@@ -0,0 +1,210 @@
+/* stylelint-disable at-rule-blacklist */
+/* stylelint-disable max-nesting-depth */
+
+// Mood Rating
+//
+// JavaScript is required for:
+// - Adding `sb-mood-rating--selected` class after value is selected, so it is possible to set previous siblings opacity.
+// - When using `sb-mood-rating--custom` or `sb-mood-rating--progressive`, SVG is hidden.
+// - For focus reasons when using `sb-mood-rating--progressive` background-position is very specific.
+//
```
beerecca: Can we add a short explanation here for the backwards ordering? Something like `Note that the reverse order of the mood rating element markup enables css-only styling of progressive hover without having to use flexbox (which doesn't meet our browser support requirements).`

src/components/_tooltip.scss
```diff
@@ -18,6 +18,7 @@
 //   </div>
 // </div>
 //
+// .sb-tooltip--upper - Slides slightly more to the top.
```
beerecca: :( Do we absolutely need this?

src/tools/_shared.scss
```diff
@@ -27,12 +27,14 @@
 
 // hide input
 @mixin sb-faux-input() {
-  appearance: none;
-  vertical-align: middle;
-  position: absolute;
   border: 0;
-  width: 0;
-  height: 0;
+  clip: rect(0 0 0 0);
```
beerecca: I'm assuming this has been checked everywhere it's used currently? Cross-browser?

src/components/_icons.scss
```diff
@@ -40,71 +40,100 @@
...
 // </svg>
-// <svg class="sb-icon sb-icon--m" viewBox="0 0 64 64">
-//   <path fill="#f4d055" d="M32 0C14.4 0 0 14.4 0 32s14.4 32 32 32 32-14.4 32-32S49.6 0 32 0z" />
-//   <path fill="#434343" d="M52 18.8c-.4-.4-.8-.8-1.2-.4l-4 1.6-10 4.8c-.4 0-.8.4-.8.8s.4 1.2.8 1.2h.4c1.2 2 2.8 3.2 5.2 3.2 3.2 0 6-2.8 6-6.4 0-.8 0-1.2-.4-2l3.6-1.6c.4-.4.4-.8.4-1.2zM12 18.8c.4-.4.8-.8 1.2-.4l4 1.6 10.4 4.8s.4.4.4.8-.4 1.2-1.2 1.2h-.4c-.8 2-2.4 3.2-4.8 3.2-3.2 0-6-2.8-6-6.4 0-.8 0-1.2.4-2L12.4 20c-.4-.4-.4-.8-.4-1.2zM46 51.6c-1.2 0-2-.8-2-2 0-7.2-9.6-8.4-12-8.4-1.2 0-12 .4-12 8.4 0 1.2-.8 2-2 2s-2-.8-2-2c0-8.8 8-12.4 16-12.4 5.2 0 16 3.2 16 12.4 0 1.2-.8 2-2 2z" />
+// <svg class="sb-icon sb-icon--m" viewBox="0 0 30 30">
```
beerecca: As discussed, we need both the old and new smilies.

src/components/_mood-rating.scss
```diff
@@ -0,0 +1,210 @@
+/* stylelint-disable at-rule-blacklist */
+/* stylelint-disable max-nesting-depth */
+
+// Mood Rating
```
beerecca: Have you noticed there's a weird re-focussing on the smiley tooltip text? Doesn't look so great :(

src/components/_mood-rating.scss
keepitterron: can we change this URL (and the one below)?
if this is part of the base assets, we need better names and probably a different location so that there is no confusion as what they are. (also, the link below refers to staging)

src/components/_tooltip.scss
```diff
@@ -44,6 +45,10 @@ $tp: #{$ns}tooltip;
   color: $sb-color-dark-3;
   transition: opacity $sb-timing-s ease-out, transform $sb-timing-m $sb-easeup-function;
 
+  &--upper {
```
keepitterron: I prefer `top`. as we already have `bottom`, `left`, `right`.

src/tools/_shared.scss
keepitterron: we repeat this bit in the `_helpers`. let's extract it to a mixin or something so we can re-use :)

src/components/_tooltip.scss
dine: You can ask @galagmon, the thing is when we don't have it the tooltip balloon does not go up enough and get in front of the smilies foreheads.

src/tools/_shared.scss
dine: Yes, that's why I updated the tooltip browser support too (it was not because of my new class)

src/components/_tooltip.scss
beerecca: What I meant sorry was can't we make this the default? Looking at the tooltip component, having a bit of a gap couldn't hurt?

src/components/_mood-rating.scss
dine: Added a [couple](https://css-tricks.com/forums/topic/transforms-cause-font-smoothing-weirdness-in-webkit/) [of solutions](http://www.useragentman.com/blog/2014/05/04/fixing-typography-inside-of-2-d-css-transforms/) for multiple browsers. Now it looks aways a little bit blurry now, but at least is consistent 😄

PR #159, opened 2018-02-15
Add LTR support for some components

clenemt:
Related to: [MALL-531](https://usabilla.atlassian.net/browse/MALL-531)

Also fix bug:
* rating component: start and end rating are not clickable when text are above


src/components/_tags.scss
```diff
@@ -58,6 +58,11 @@
     margin: sb-px2rems(3px);
     margin-left: 0;
     padding: sb-rem-spacing('-xs') sb-rem-spacing('');
+
+    .#{$ns}rtl {
```
dine: `.#{$ns}rtl &`

src/components/_dropdown.scss
```diff
@@ -1,3 +1,5 @@
+/* stylelint-disable max-nesting-depth */
```
dine: It would be nice if we didn't have to disable this.

keepitterron: agreed.
in this case, tho, the cons of not disabling the rule are not negligible (`.#{$ns}rtl .#{$ns}dropdown__button .#{$ns}input` is hard to read and reason about). but disabling rules can become a nasty habit.

not sure how can we elegantly solve this.

docs-builder/kss-assets/kss.js
```diff
@@ -84,6 +84,15 @@ var logo = require('./logo.svg');
   addSectionSeparator(sectionsDepth2);
 })();
 
+// Allow for testing components in RTL by pressing the 'R' key
+(function () {
+  document.addEventListener('keypress', function(event) {
+    if (String.fromCharCode(event.charCode).toUpperCase() !== 'R') return;
```
keepitterron: 🎉

src/components/_dropdown.scss
dine: Considering the legibility in this case, I'd say we disable and be careful not to overuse it?

beerecca: We only need to remove one level though right? We could have something like: 
```
.#{$ns}dropdown {
   @include sb-text-s();
   position: relative;
   color: $sb-color-dark-3;
}

.#{$ns}dropdown__button {
     z-index: $sb-z-btn;
 		 
     .#{$ns}input {
       padding-right: sb-px2rems(30px);
       cursor: pointer;
       text-align: left;
        overflow: hidden;
        text-overflow: ellipsis;
        white-space: nowrap;
 +
 +      .#{$ns}rtl & {
 +        text-align: right;
 +        padding-right: sb-px2rems(11px);
 +        padding-left: sb-px2rems(30px);
 +      }
      }		     
 }
```

I'd much prefer to add a bit of verbosity (as long as it's still easy to reason about, I agree there) than to ignore the rule.

src/components/_dropdown.scss
```diff
@@ -43,6 +45,12 @@
       overflow: hidden;
       text-overflow: ellipsis;
       white-space: nowrap;
+
+      .#{$ns}rtl & {
+        text-align: right;
+        padding-right: sb-px2rems(11px);
```
beerecca: Would be nice to use sb-rem-spacing function - I know it's not used everywhere, but it's less likely we're gonna make up our own numbers the more we use it.

clenemt: this was just to use the exact same one as the ltr uses, should I change them all?

beerecca: If it's not too much work that would be nice!

clenemt: Agreed, then let's merge this PR and i will do it inside another. Ok ?

src/components/_dropdown.scss
clenemt: The rule was already disabled in two files already, that is why I thought it was okay.
I think the nesting problems comes partly from how we write our BEM components.
Here is a pick from 2014: https://www.sitepoint.com/beware-selector-nesting-sass/#unsearchable-selectors

Also for the dropdown it works fine but for some other components not so much (e.g. moodrating)

beerecca:
Approving on the understanding that sb-rem-spacing and nesting will be done in another pr.

PR #160, opened 2018-02-15
Add LTR support for checkboxes and radios

clenemt:
Related to: [MALL-531](https://usabilla.atlassian.net/browse/MALL-531)

**BREAKING CHANGE:**

In order to support RTL we have to move away from utility classes as base markup.

They are currently used to add spacing between checkboxes/radios and their (text) label.

This commit move the util class as part of the css and create a new class for the (text) label.

Before and after:

```
<span class="sb-margin-left-xs">Basic checkbox</span>
<span class="sb-checkbox__text">Basic checkbox</span>
```


src/components/_checkbox.scss
```diff
@@ -65,6 +65,15 @@
     transition: transform $sb-transition;
   }
 
+  &__text {
+    margin-left: sb-rem-spacing('-xs');
```
dine: I think we can add this aldo to the `_shared.scss`, because we do the same in radio and checkbox

PR #166, opened 2018-02-22
fix: move away from SASS nesting as much as possible

clenemt:
This makes it:
* easy on the eyes when reading stylabilla from code or source maps
* more searchable

https://www.sitepoint.com/beware-selector-nesting-sass/


src/components/_dropdown.scss
```diff
@@ -34,102 +32,102 @@
...
-      display: inline;
-      pointer-events: none;
-    }
+.#{$ns}dropdown__button {
+  z-index: $sb-z-btn;
 
-    .#{$ns}input:disabled ~ .#{$ns}icon {
-      fill: $sb-color-light-1;
-    }
+  &:disabled ~ .#{$ns}icon {
```
beerecca: Doesn't this need to be `.#{$ns}input:disabled`?

clenemt:
Updated and ready for review

clenemt:
I get your points @keepitterron, but if I may add:

I searched for stuff in the DOM/codebase way more often then I rename stuff so I think it is fine (just did in `usabilla.js` and got false positive searching for a class name).
I tried not to remove nesting from components, just the BEM part of it.

PR #175, opened 2018-04-17
fix(mood-rating): allow correct keyboard navigation

tzengerink:
BREAKING CHANGE: Due to the reversed element order, the mood rating
could not be navigated in an accessible way. The reversed order was
necessary to be able to apply CSS-only styling to the progressive mood
rating.

This commit uses the proper HTML order again and fixes the issue by
using the `sb-mood-rating--selected` class. This solution aligns with
the way the opacity is solved for the default and custom mood ratings.

See: https://usabilla.atlassian.net/browse/BUB-33
Reviewers: @beerecca @dine @clenemt @spirosikmd 

gxbybofpoi


src/components/_mood-rating.scss
```diff
@@ -6,39 +6,47 @@
 // - Adding `sb-mood-rating--selected` class after value is selected, so it is possible to set previous siblings opacity.
```
dine: Let's change `...set previous siblings opacity` to `...set previous siblings styles`

src/components/_mood-rating.scss
```diff
@@ -102,7 +101,7 @@
 //
 // .sb-mood-rating--default - Default SVG showing rating
 // .sb-mood-rating--custom - Shows background image instead of SVG
-// .sb-mood-rating--progressive - Progressive mood rating, does not show tooltips or SVG
+// .sb-mood-rating--progressive.sb-mood-rating--selected - Progressive mood rating, does not show tooltips or SVG
```
dine: Should we add a `// .sb-mood-rating--progressive - Progressive mood rating, no value selected` here? 
Or maybe remove the first 2 examples without `.sb-mood-rating--selected`? 
I'd say add and then move this line to be the last one (keeping same order before value is selected and after).

PR #176, opened 2018-04-18
fix(mood-rating): apply hover styles before selected

dine:
Before:

After:


spirosikmd:
@dine I am noticing a slight move to the top in the `After` case. Why is this happening?

dine:
@spirosikmd I think I scrolled the page a little bit while recording the gif. It doesn't actually happen.

PR #179, opened 2018-05-07
fix(rating): add inline rating rtl support

dine:
Fix order of low/high labels for inline rating when using direction rtl.
To test click 'r' after loading page.

#### Before
screen shot 2018-05-07 at 14 15 02

#### After
screen shot 2018-05-07 at 14 16 16


dine:
Yep, this is IE10 🙂 

screen shot 2018-05-08 at 09 58 04

PR #185, opened 2018-05-14
fix(style): updates to match designs

dine:
- Update close icon;
- Remove drop shadow for all links focus.

Before:
screen shot 2018-05-14 at 17 14 22

After:
screen shot 2018-05-14 at 17 14 55


kagnamazian:
@dine let me know if the production deploy crash. Since we create the `latest` bucket on S3 want to be sure that it will replace it and not crashed since it already exist (should not crash normally ^^)

PR #187, opened 2018-07-02
feat(form-control): Add icon support for form-control

clenemt:
:arrow_forward: **Related to**: BLOS-83
:eyeglasses: **Context**: Adds support for left and right icons inside `.form-control` inputs with the two new classes `.form-control__input--icon-left` and `.form-control__input--icon-right`. This is a **breaking change**, since the dropdowns and selects now need a new class added to them to work like before.

:sunrise: **After**:


src/components/_form-control.scss
```diff
@@ -37,18 +37,63 @@
...
   .#{$ns}icon {
     position: absolute;
     top: 50%;
-    right: sb-px2rems(7px);
     transform: translateY(-50%);
+    left: auto;
+    right: sb-rem-spacing('s');
+  }
+
+  .sb-input {
```
spirosikmd: We should use the namespace identifier `#{$ns}` instead of `sb` in all these places, except when calling the functions.

src/components/_form-control.scss
```diff
@@ -37,18 +37,63 @@
...
+    padding-left: sb-rem-spacing('s');
+    padding-right: sb-px2rems(34px);
```
spirosikmd: Why aren't we using a specified size here instead of this hardcoded `34px`?

clenemt: Because padding left and right is 11 plus 16px for the icon size = 38px but for some reaseon it looks better with 34px. Will show @galagmon

spirosikmd:
Without an icon, the input should have `padding-right: sb-rem-spacing('s')`, right?

screen shot 2018-07-03 at 09 56 23

clenemt:
I fix this without adding a breaking change, but the code becomes a bit of a mess. I would prefer adding a breaking change, removing the not needed `.sb-form-control__input` and adding the classes on the `.form-control--icon-left` and `.form-control--icon-right`. Let me know.

spirosikmd:
This is what we have now:
screen shot 2018-07-04 at 09 41 56
I am ok with breaking the backwards compatibility as long as there isn't any other way to fix this. I like the approach with the two explicit classes for right and left icon. I don't think that we can remove the `.sb-form-control__input` class, as it is now used in the `dropdown` and `select` components.

clenemt:
So just to clarify (as I am still waiting for other reviewers), you are in favor of adding `.form-control--icon-left` and `.form-control--icon-right` instead of a single `.form-control__input--left` that this PR is adding. Am I correct?

galagmon:
@clenemt can we somehow align the icon to the center of the input field and not the text bottom?

clenemt:
@galagmon the icon is actually centered to the input. This is mostly an issue with the search icon that does not fill the 16 by 16 as shown in the screenshots below:

screen shot 2018-07-11 at 10 48 35

It looks better with the star for instance: 

screen shot 2018-07-11 at 10 50 21

clenemt:
@spirosikmd breaking changes breaks the dropdown: 

screen shot 2018-07-11 at 10 56 25

As it no longer aligns the icon right by default. Are you ok with that?

spirosikmd:
I am ok with fixing it of course 😄 So to summarise my view on this:

- I prefer to fix the current issue with the text overlapping with the right icon.
- I prefer to use two explicit classes for icon left and icon right. The advantage of this is that it doesn't depend on a specific DOM structure.
- These changes will break the `dropdown` and `select` components, so we should update these components to reflect the new changes.
- I am fine with this being a breaking change.

Does this make sense?

src/components/_form-control.scss
```diff
@@ -37,18 +37,68 @@
...
 
     .#{$ns}rtl & {
       right: auto;
-      left: sb-px2rems(7px);
+      left: sb-rem-spacing('s');
     }
   }
 }
+
+.#{$ns}form-control__input--icon-right .#{$ns}input {
```
dine: Can't we combine this rules with the on line 94, like we did below with `...-left` + `...rtl ...-right`?
https://github.com/usabilla/stylabilla/pull/187/files#diff-a723c6508c22eb0b44884ed6234e9fddR94

clenemt:
Updated the PR code and description. Have a look please  🙏

PR #188, opened 2018-07-16
feat(grid): add large gutter as grid modifier class

clenemt:
:arrow_forward: **Related to**: BLOS-110
:eyeglasses: **Context**: we need a larger gutter value for the analysis-app when displaying the list of workspaces. This PR adds `.sb-grid--gutter-l` as a new grid modifier class.

:sunrise: **After**:

screen shot 2018-07-16 at 12 57 59


src/layout/_layout.scss
```diff
@@ -11,7 +11,7 @@
 
 $sb-grid-columns: 12;
 $sb-default-gutter: 16px;
-$sb-gutter-map: ('-0': 0, '-s': 11px);
+$sb-gutter-map: ('-0': 0, '-s': 11px, '-l': 23px);
```
clenemt: **Side note**: because we are using `width: calc( col% - gutter)` for the columns value, adding a new gutter creates a lot of css (e.g. we need to rewrite all those `width: calc()` for every breakpoints, cols & offsets). This can be visualized when looking at the current css output.

Another option would be to move to fixed `width: col%` with margin and padding for all gutters. This option creates less code for our usecase.

PR #189, opened 2018-07-17
 feat(sidepanel): add sidepanel component

dine:
🖼 
localhost_8080_section-components html
localhost_8080_section-components html 1

🎥


src/components/_sidepanel.scss
```diff
@@ -0,0 +1,66 @@
...
+
+      @include sb-focus-no-border();
+    }
+  }
+}
+
+.#{$ns}sidepanel--collapsed {
+  .#{$ns}sidepanel__container {
+    padding: 0;
+    width: 0;
```
clenemt: Not sure that the width animation is what you want as everything kinds of shrinks down (even more visible when lots of content). A translation might be a better way to achieve that animation.

src/components/_sidepanel.scss
```diff
@@ -0,0 +1,66 @@
+// Sidepanel
+//
+// Browser support: IE9+ (no transition), Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+// <aside class="sb-sidepanel {{modifier_class}}" role="menu" aria-expanded="true">
+//   <div class="sb-sidepanel__container">
+//     <div class="sb-padding-l">
```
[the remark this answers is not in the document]
dine: When collapsing/opening it gets a weird animation if we have padding and transition the width... Also, right now we'll just have a button inside, but I think we'll need this wrapper in the future for scrolling purposes too, if we have too much content.

galagmon:
@dine did we add a transition to flip the arrow?

dine:
@galagmon yep, I'll add gifs here after addressing @clenemt comments

src/settings/_box.scss
```diff
@@ -23,3 +23,6 @@ $sb-mood-size: 30px;
 // Tooltip arrow size and shadow container
 $sb-tooltip-arrow-size: 11px;
 $sb-tooltip-arrow-container: 26px;
+
+// Main nav & sidepanel
```
spirosikmd: Is this just the side panel width? Or does it contain the main nav width as well? It should only be about the side panel, right?

src/components/_icons.scss
```diff
@@ -101,6 +101,13 @@
 // <svg class="sb-icon" viewBox="0 0 16 16">
 //   <path d="M6.5 12a5.5 5.5 0 1 0 0-11 5.5 5.5 0 0 0 0 11zm0 1a6.5 6.5 0 1 1 0-13 6.5 6.5 0 0 1 0 13zM7 6V3H6v3H3v1h3v3h1V7h3V6H7z" />
 // </svg>
+// <svg class="sb-icon sb-icon--handle" viewBox="0 0 23 23">
+//   <g fill="none" fill-rule="evenodd">
+//     <circle cx="11.5" cy="11.5" r="11.5" fill="#FFF"/>
+//     <path fill="#59636B" fill-rule="nonzero" d="M10.52 12.55l1.62 1.65c.4.4.4 1.07 0 1.49a1 1 0 0 1-1.45 0L7.3 12.25c-.4-.41-.4-1.09 0-1.5l3.4-3.44a1 1 0 0 1 1.44 0c.4.42.4 1.08 0 1.5l-1.62 1.64h4.45c.57 0 1.03.47 1.03 1.05 0 .58-.46 1.05-1.03 1.05h-4.45z"/>
```
beerecca: Are these stylabilla colours? This is inconsistent with how we do `sb-icon--mood__lines` for example.

src/components/_icons.scss
```diff
@@ -158,6 +165,11 @@
   fill: $sb-color-mood-heart-eyes;
 }
 
+.#{$ns}icon--handle {
+  height: sb-px2rems(23px);
```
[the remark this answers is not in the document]
beerecca: Yea we made some not great decisions here that are now biting us in the butt.
sb-spacing-map has `'m': 16px, 'l': 23px, 'xl': 40px`. Which it looks like we want all of those sizes now, plus mood rating being 30. Assuming that you've checked this icon at 16px and it's really unusable?

Ideally we'd have $sb-icon-size-map: ('--l': 23px, '--xl': 40px);
And hard code a new `sb-icon--mood` class that uses $sb-mood-size (30px) seeing as that's the unique case that doesn't fit into our scale.

I don't really see a way of making this work (ie coming up with a size between s and m) without breaking changes so I think we just need to do it.

src/components/_icons.scss
beerecca: Also have you checked this svg against https://jakearchibald.github.io/svgomg/ or something? I think there is some unnecessary metadata in here.

src/components/_sidepanel.scss
```diff
@@ -0,0 +1,67 @@
...
+    left: sb-px2rems($sb-sidepanel-width);
+    transition: left $sb-timing-s ease-in-out, transform $sb-timing-m ease-in-out;
+    transform-origin: 50% 50%;
+    transform: translate(-50%, -50%);
+    z-index: $sb-z-header;
+
+    &:focus {
+      border-radius: 50%;
+
+      @include sb-focus-no-border();
```
beerecca: Is this ok accessibility wise? You should still be able to see you're focussed on it right?

src/components/_sidepanel.scss
```diff
@@ -0,0 +1,67 @@
...
+    border: 0;
+    padding: 0;
+    cursor: pointer;
+    position: absolute;
+    top: 50%;
+    left: sb-px2rems($sb-sidepanel-width);
+    transition: left $sb-timing-s ease-in-out, transform $sb-timing-m ease-in-out;
+    transform-origin: 50% 50%;
+    transform: translate(-50%, -50%);
+    z-index: $sb-z-header;
```
spirosikmd: Should we use a specific `z-index` for this handle? It now has a higher `z-index` than the drop down. I am not sure if we see a drop down in the future displaying under this handle.

src/components/_icons.scss
```diff
@@ -158,6 +165,11 @@
   fill: $sb-color-mood-heart-eyes;
 }
 
+.#{$ns}icon--handle {
```
spirosikmd: Is this arrow really a `handle` icon? Shouldn't we use something like left or right arrow?

src/components/_icons.scss
dine: @beerecca I did, this is the optimised version. Considering the new designs I'll export a new svg just with the arrow and add the border to the button (using css).

src/components/_icons.scss
dine: @spirosikmd I'll replace it and rename it to arrow.

src/components/_icons.scss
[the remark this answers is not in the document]
dine: Only the arrow icon is currently 8px so we get a new size anyway. We would have to choose between `'xs': 7px` or `'s': 11px`. @michelvdhoek can we use our scale here? If yes, would you prefer 1px smaller or 3px bigger?

src/components/_sidepanel.scss
dine: I think we'll need something here. Only the handle has to go partially on top of the main navigation, while the sidepanel goes below. It doesn't use too much space, so we shouldn't have a dropdown issue.

src/components/_icons.scss
```diff
@@ -158,10 +164,17 @@
   fill: $sb-color-mood-heart-eyes;
 }
 
-$sb-icon-size-map: ('--m': $sb-mood-size, '--l': 40px);
+.#{$ns}icon--arrow {
+  stroke: $sb-color-dark-3;
```
spirosikmd: Why are we giving a specific stroke color for the arrow icon? Can't this be the default text color (as with other icons) and change it in case we need it to?

beerecca: Tbh @spirosikmd from what Gal and Michel are saying, these icons are going to change when Marco starts as one of his first tasks is a consistent icon library (as you can probably tell, our icons all look a bit different right now). There are other problems with naming (eg `arrow`, `arrow-left`) which also are not ideal. So personally I think this is fine for now considering it will all change.

PR #191, opened 2018-07-19
feat(modal): add modal

clenemt:
:arrow_forward: **Related to**: BLOS-113
:eyeglasses: **Context**: adds the modal component so we can use it as a react component later on.

:camera_flash: **After**:

screen shot 2018-07-19 at 12 23 22


src/components/_icons.scss
```diff
@@ -33,7 +33,7 @@
 // <svg class="sb-icon" viewBox="0 0 16 16">
 //   <path d="M7.83 9.243c-1.943 1.286-4.584 1.074-6.294-.636-1.953-1.953-1.953-5.12 0-7.07 1.952-1.954 5.118-1.954 7.07 0 1.71 1.71 1.923 4.35.637 6.292l2.906 2.904c.386.387.385 1.015-.008 1.408-.39.39-1.026.388-1.408.007L7.83 9.242zM7.9 7.9c1.562-1.563 1.562-4.095 0-5.657C6.336.68 3.804.68 2.242 2.243.68 3.805.68 6.337 2.243 7.9 3.805 9.46 6.337 9.46 7.9 7.9z"/>
 // </svg>
-// <button class="sb-icon sb-icon--close sb-icon--l" aria-label="Close">&times;</button>
+// <button class="sb-btn sb-btn--subtle" aria-label="Close"><span class="sb-icon sb-icon--close" aria-hidden="true">&times;</span></button>
```
clenemt: **Big problem currently:** The current button height is based on the text height. Because I bump the font-size to be able to see the `&times;`, the button is now bigger then the others buttons 🤢

Two additional options for adding the close icons:
* Add an SVG icon instead of the current span with `&times;` in it. This would remove the need to reset all those styles.
* Not use the icon classes for that and create an `sb-btn sb-btn--close` (but we still have the button height issue)

src/components/_modal.scss
```diff
@@ -0,0 +1,81 @@
...
+.sb-modal__dialog {
+  width: 500px;
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  opacity: 0;
+  transform: translateY(-40%) translateX(-50%);
+  transition: transform $sb-timing-m $sb-easing-function, opacity $sb-timing-s $sb-easing-function;
+
+  .sb-modal--open & {
```
clenemt: Need to work a bit more on the animation classes as it is very dependant on the solution we pick for `react-stylabilla`

src/components/_icons.scss
spirosikmd: As far as I understood, the problem is that we use the close button in the form app and we cannot update it as we are going to break the compatibility? 

If that is the case, is it possible to add another SVG icon (as you proposed) that we can then use with the button and deprecate the old one? We can then change the form to use the new one later.

src/components/_modal.scss
```diff
@@ -0,0 +1,91 @@
+// Modal
+//
+// A modal component showing fullscreen.
+//
+// Browser support: IE9+ (no transition), Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+// <div class="sb-modal sb-modal--open" role="dialog">
+//   <div class="sb-modal__backdrop" aria-hidden="true"></div>
+//   <div class="sb-modal__dialog" role="document">
```
spirosikmd: What is this `role="document"`? According to [this](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_dialog_role) article, dialogs should have proper labeling with `aria-labelledby` and `aria-describedby`.

src/components/_modal.scss
```diff
@@ -0,0 +1,91 @@
...
+//         <button class="sb-btn sb-btn--subtle sb-margin-right">Cancel</button>
+//         <button class="sb-btn sb-btn--primary">OK button</button>
+//       </div>
+//     </div>
+//   </div>
+// </div>
+//
+// Styleguide Components.modal
+
+.sb-modal {
```
spirosikmd: We should use the `.#{$ns}` in this file!

src/settings/_z-index.scss
```diff
@@ -2,6 +2,7 @@ $sb-z-btn-disabled: 0;
 $sb-z-btn: 10;
 $sb-z-btn-active: 20;
 $sb-z-btn-hover: 30;
+$sb-z-modal: 50;
```
spirosikmd: How did we come up with this z-index? Why is it less than the dropdown and the header?

src/components/_modal.scss
clenemt: role document is the one used by most frameworks out there for good reasons, see bootstrap, & materialui. I will add the aria though, good point.

src/components/_icons.scss
beerecca: Just saw this re the form app - there is no issue with breaking compatibility there, do whatever you like as this project is not live and probably won't be for quite a while. But we did discuss this with Gal a couple of weeks ago, I thought we came to a conclusion...

src/settings/_z-index.scss
clenemt: A better way to put it is: "can you explain why the z-index is 50?" 😉 
I did it this way because we want modal to hold dropdowns in it.
But as a sidenote, since we are not moving dropdowns to document body and they are not position fixed, it does not really matter.

The only clash will be with fixed element, which we only have toast for now.
So maybe put modal above toast as 300?

src/components/_icons.scss
clenemt: We did came with the solution implemented in this PR. I am just pointing out the problems of that solution...

spirosikmd: Ok then if we all agree on this solution and we are all aware of its tradeoffs then let's go for it!

src/components/_modal.scss
```diff
@@ -0,0 +1,91 @@
...
+  position: fixed;
+  z-index: -1;
+  width: 100%;
+  height: 100%;
+  top: 0;
+  left: 0;
+  background-color: $sb-color-dark-2;
+  opacity: 0;
+  transition: opacity $sb-timing-s ease-out;
+  -webkit-tap-highlight-color: transparent;
```
spirosikmd: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-tap-highlight-color) this is a non-standard CSS property and we shouldn ot use it in production. What was the reason to use it here?

clenemt: It prevents black-highlights when taping the background on ios

spirosikmd: Since we don't have iOS users for the app (at the moment), and since this is not recommended for production, maybe we should avoid using it?

clenemt: We are supporting tablets though, which means we are supporting iOS.
Also quoting that MDN article: "it will not work for every user", of course because we are only removing the grey backdrop on iOS devices.

spirosikmd:
I would love @galagmon and @beerecca review on this one!

PR #193, opened 2018-07-24
[BUB-159] Add overflow menu

tzengerink:
Part of: BUB-159
Reviewers: @dine @lcobucci @tzengerink @gPinato @dsantang @beerecca @galagmon @spirosikmd


src/components/_overflow-menu.scss
```diff
@@ -0,0 +1,78 @@
+// Overflow Menu
```
spirosikmd: It would be nice to override the styles of kss to set a fixed height like we did for the dropdown. You probably need to override `kssref-components-overflow-menu .kss-modifier__wrapper` and set a height.

src/components/_overflow-menu.scss
```diff
@@ -0,0 +1,78 @@
...
+//
+// Browser support: IE9+ (no transition), Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+// <div class="sb-overflow-menu">
+//   <div class="sb-overflow-menu__item">
```
spirosikmd: For the dropdown we use a `ul` with `li` as the items. Why have we decided to use `div`s here?

src/components/_overflow-menu.scss
```diff
@@ -0,0 +1,78 @@
+// Overflow Menu
+//
+// Browser support: IE9+ (no transition), Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+// <div class="sb-overflow-menu">
```
spirosikmd: How about accessibility? Have we thought about it? For the dropdown we use `role="listbox"` in the menu and `role="option"` for the items. We should also make it keyboard navigatable, right?

src/components/_overflow-menu.scss
```diff
@@ -0,0 +1,78 @@
...
+  cursor: pointer;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+
+  .#{$ns}icon {
+    margin-right: sb-px2rems(3px);
+  }
+
+  span {
```
spirosikmd: Let's use a class here and not depend on specific HTML elements. Maybe `sb-overflow-menu__item-label` or `overflow-menu__item-text`?

src/components/_overflow-menu.scss
```diff
@@ -0,0 +1,83 @@
...
+//   </li>
+// </ul>
+//
+// Styleguide Components.overflow-menu
+
+.#{$ns}overflow-menu {
+  @include sb-text-s();
+
+  position: absolute;
+  width: 140px;
```
beerecca: This smells like a magic number to me :D Should probably at least be in a variable, maybe declared alongside the other variables like header height.

src/components/_overflow-menu.scss
```diff
@@ -0,0 +1,83 @@
...
+  border-radius: sb-px2rems($sb-radius);
+  box-shadow: $sb-box-shadow;
+  z-index: $sb-z-dropdown;
+}
+
+.#{$ns}overflow-menu__item {
+  padding: sb-rem-spacing('s');
+  cursor: pointer;
+  overflow: hidden;
+  text-overflow: ellipsis;
```
beerecca: Is this confirmed with design, what they want to happen when the text is too long for the 140px?

src/components/_overflow-menu.scss
```diff
@@ -0,0 +1,83 @@
...
+  white-space: nowrap;
+
+  .#{$ns}icon {
+    margin-right: sb-px2rems(3px);
```
beerecca: Another magic number, do we not have any existing variables that give us close enough to what we need?

src/components/_overflow-menu.scss
```diff
@@ -0,0 +1,83 @@
...
+    color: $sb-color-red-2;
+
+    &:active {
+      color: $sb-color-white;
+      background-color: $sb-color-red-2;
+    }
+  }
+}
+
+.#{$ns}overflow-menu__item-label {
```
beerecca: This isn't strictly BEM - I'd prefer to flatten the grandchild (https://assortment.io/posts/grandchild-elements-bem-css) and call it `overflow-menu__label`, I think this is the approach we take elsewhere in stylabilla.

src/index.scss
```diff
@@ -40,30 +40,31 @@
 @import 'layout/flex';
 
 // Components
-@import 'components/tile';
-@import 'components/checkbox';
-@import 'components/radio';
-@import 'components/input';
+@import 'components/alert';
```
beerecca: So ideally this would be fine, components should not rely on the cascade between them, but have you tested that everything still works? For instance alert might need animation to be first?

beerecca: True. I'd suggest testing this Teun and then moving anything that is relied upon.

src/components/_overflow-menu.scss
tzengerink: I've now made the width of the menu depend on the length/content of the menu items. If however a fixed with is set that is smaller, we should use ellipsis to cut of the content. This is how I discussed it with design.

src/components/_overflow-menu.scss
```diff
@@ -0,0 +1,92 @@
...
+  margin: 0;
+  padding: 0;
+  background-color: $sb-color-white;
+  border: 1px solid $sb-color-light-3;
+  border-radius: sb-px2rems($sb-radius);
+  box-shadow: $sb-box-shadow;
+  z-index: $sb-z-dropdown;
+
+  &--open {
+    display: block;
```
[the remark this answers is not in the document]
tzengerink: I like the animation of the tooltip, the dropdown menu however does not have an animation. Since this is also kind of a menu, I think we should be consistent in the use of animations.

src/components/_overflow-menu.scss
```diff
@@ -0,0 +1,110 @@
+// Overflow Menu
+//
+// Browser support: IE9+, Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+// <ul class="sb-overflow-menu sb-overflow-menu--open" tabindex="0" role="listbox">
+//   <li class="sb-overflow-menu__header">
+//     user@usabilla.com
+//   </li>
```
tzengerink: Having some doubts about this not being really semantic. Other solutions would be to wrap the entire element in a `div` to allow for an extra `div` above the `ul` to contain the header. I choose this solution, because if the menu is used without a header it would have to be wrapped in a `div`, while right now it is just a `ul` with child `li` elements. Ideas?

src/components/_overflow-menu.scss
```diff
@@ -0,0 +1,110 @@
+// Overflow Menu
+//
+// Browser support: IE9+, Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+// <ul class="sb-overflow-menu sb-overflow-menu--open" tabindex="0" role="listbox">
```
clenemt: I arrive a bit late to the discussion, but I think this should be a `role="menu"` with `role="menuitem"`. See: https://www.w3.org/TR/wai-aria-practices-1.1/#menu for differences.

src/components/_overflow-menu.scss
```diff
@@ -0,0 +1,110 @@
...
+}
+
+.#{$ns}overflow-menu__header,
+.#{$ns}overflow-menu__item {
+  padding: sb-rem-spacing('s');
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+
+  .#{$ns}icon {
```
clenemt: I'm inclined to simply add a util class to the icon instead of overriding a style e.g. `sb-margin-right-xs`, what do you think?

src/components/_overflow-menu.scss
```diff
@@ -0,0 +1,110 @@
...
+  line-height: sb-px2rems(23px);
+  font-weight: 600;
+}
+
+.#{$ns}overflow-menu__item {
+  cursor: pointer;
+
+  &:hover,
+  &--highlighted {
+    background-color: $sb-color-light-4;
```
clenemt: Should we use a different color for hover and selected @galagmon ?

src/components/_overflow-menu.scss
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  &:active {
```
clenemt: `active` state is only for buttons and links (https://html.spec.whatwg.org/multipage/semantics-other.html#selector-active). Maybe add a `menu__item` with either button or links to your example?

src/components/_overflow-menu.scss
```diff
@@ -0,0 +1,110 @@
...
+  .#{$ns}icon {
+    margin-right: sb-rem-spacing('xxs');
+  }
+}
+
+.#{$ns}overflow-menu__header {
+  font-family: $sb-primary-font-family;
+  font-size: sb-px2rems(16px);
+  line-height: sb-px2rems(23px);
+  font-weight: 600;
```
clenemt: It got moved to `500` for bold now on `master` branch

src/components/_overflow-menu.scss
```diff
@@ -0,0 +1,110 @@
...
+  text-overflow: ellipsis;
+  white-space: nowrap;
+
+  .#{$ns}icon {
+    margin-right: sb-rem-spacing('xxs');
+  }
+}
+
+.#{$ns}overflow-menu__header {
+  font-family: $sb-primary-font-family;
```
clenemt: This seems very similar to `sb_text`, maybe remove those styles and add a `sb_text` class to the example? This would make sure that they stay in synch if we ever change how text looks like.

src/components/_overflow-menu.scss
```diff
@@ -0,0 +1,110 @@
...
+
+  &:active {
+    background-color: $sb-color-blue-4;
+  }
+
+  &:focus {
+    outline: none;
+  }
+
+  &--warning {
```
clenemt: Based on the color you are using, it should be `&--danger`. This would be consistent with `.sb-btn--danger` for instance.

src/components/_overflow-menu.scss
dine: Maybe test it before changing if it is the desired behaviour: http://adrianroselli.com/2017/10/dont-use-aria-menu-roles-for-site-nav.html

src/components/_overflow-menu.scss
```diff
@@ -0,0 +1,110 @@
...
+//      </svg>
+//      <span class="sb-overflow-menu__label">Delete</span>
+//   </li>
+// </ul>
+//
+// Styleguide Components.overflow-menu
+
+.#{$ns}overflow-menu {
```
clenemt: I missed that part, but looking at the PR for react-stylabilla, should this component not have a `min-width`?
It looks a bit cramped otherwise.

clenemt: Also @galagmon confirmed that it should have a `min-width: 150px;`

src/components/_overflow-menu.scss
tzengerink: This is similar to how the Dropdown is implemented, which was what @spirosikmd suggested?

src/components/_overflow-menu.scss
tzengerink: I like the idea of not cluttering the HTML with to much classes. Also makes implementing this style in components a bit easier.

src/components/_overflow-menu.scss
clenemt: Sorry what? I don't understand what you mean, maybe @spirosikmd can help me?

spirosikmd: @clenemt from the spec you linked here, I can't read anywhere that `active` is only for buttons and links. More specifically it writes that:

```
If the element has its tabindex focus flag set
    The element is being activated if it is in a formal activation state.
```

which means that if we have a `li` tag with `tabindex` set (as in the case of `dropdown`) then `active` state is valid. At least this is the way I read this! Specs are hard 😄

clenemt: Nha I think the one that triggers the active state is this:

> If the element is being actively pointed at

I removed the `tabindex` and it is still clickable (showing the blue `:active`).
Nonetheless, I find it weird, maybe time will change my mind. ⌛️

src/components/_overflow-menu.scss
dine: I missed this before, but I think I agree with @clenemt here

PR #197, opened 2018-08-01
Multiple typographic enhancements

clenemt:
:arrow_forward: **Related to**: BLOS-109
:eyeglasses: **Context**: while working on the card, I found out multiple issues:

**1-  Our usage of font-weight is strange:**
* on the documentation website we forgot to include the medium font and were using faux-bold
* on stylabilla we were sometime using faux-bold in place of the medium font
* move the medium font-weight from 600 to 500 to match @galagmon designs

**2-  We are missing some common text helpers:**
* `sb-text--uppercase`
* `sb-text--lowercase`
* `sb-text--number` (needed since our milo font has number not aligned with text)

**3-  We are missing the font smoothing currently present on the app.usabilla.com:**
* This led to adjusting some styles on the documentation website to make it more readable

:sunrise: **After**:


clenemt:
Awaiting moar reviews for that one as it might have some big impacts 👍

src/elements/_page.scss
```diff
@@ -23,4 +23,6 @@ html {
   font-family: $sb-primary-font-family;
   color: $sb-color-dark-3;
   font-weight: 400;
+  -webkit-font-smoothing: antialiased;
```
dine: We had a discussion about this in the past: https://github.com/usabilla/stylabilla/pull/21#discussion_r128528569
And we ended up removing the font-smoothing. 

Maybe it's good to just do a double check with the designers in other OS/Browser combinations, since we all use Chrome/Mac. I remember it affected Windows/IE in a way that fonts would look really bad and affect readability/accessibility. 

I also found this (maybe worth a try?) 🙂  https://gist.github.com/hsleonis/55712b0eafc9b25f1944

src/trumps/_text.scss
```diff
@@ -80,3 +83,16 @@
...
+.#{$ns}text--uppercase {
+  text-transform: uppercase;
+}
+
+.#{$ns}text--lowercase {
+  text-transform: lowercase;
+}
+
+.#{$ns}text--number {
+  font-variant-numeric: lining-nums;
```
dine: I think most of the times we use Milo font we'll want this as the default, right? 
If that is the case I'd say instead of adding this class every time we need to use numbers, maybe we should add this rule to the same block where we apply Milo as the default font for `html`? https://github.com/usabilla/stylabilla/blob/2512a69b5248604201db0665ee6e0a3fa0511735/src/elements/_page.scss#L22

Also to be discussed with the designers...

docs-builder/kss-assets/kss.scss
```diff
@@ -603,6 +607,7 @@ $kss-vertical-rhythm      : $kss-font-size * 1.5;
     font-family: monospace;
     background: $kss-colors-background;
     margin-right: 10px;
+    font-weight: bold;
```
beerecca: Shouldn't this also be 500?

src/components/_button.scss
```diff
@@ -39,7 +39,7 @@ $sb-btn-padding-l: sb-rem-spacing('s') sb-rem-spacing('l');
   display: inline-block;
   font-family: $sb-primary-font-family;
   font-variant: all-small-caps;
-  font-weight: 600;
+  font-weight: 500;
```
beerecca: Are we absolutely sure on this one? Have you shown @galagmon the impact it would have? Might be nice to show her what it looks like if you change it on the button creator for example.

galagmon:
@beerecca I asked for these changes and the main objective is to have the right classes for UB next, I know it might have some impact on the component already implemented in Usabilla. Also, it is not best practice using bold instead of the actual font weight so I think it will only make things look cleaner. I would like to see the impact on the button editor if possible. (send me a screen shot)

src/elements/_page.scss
clenemt: Will try it, thanks

clenemt: Just did a few tests on windows: ie 11/ edge/firefox/chrome with browser tests, nothing really changes. Also as per caniuse: https://caniuse.com/#search=font-smooth it should work only on osx

docs-builder/kss-assets/kss.scss
clenemt: Nope since the font is monospace and a 500 in monospace just displays the regular one.

src/components/_button.scss
clenemt: Yes I have shown gal how it looks, also this does not change anything since we moved the declaration of the bold font from 600 to 500 cf: `src/settings/_fonts.scss` below.

PR #199, opened 2018-08-07
chore: update circle config

beerecca:
@spirosikmd 🤞


.circleci/config.yml
```diff
@@ -0,0 +1,81 @@
...
+          requires:
+            - test_and_npm_release
+  prod:
+    jobs:
+      - test_and_npm_release:
+          filters:
+            branches:
+              only: master
+            tags:
+              only: /v[0-9]+(\.[0-9]+)*/
```
spirosikmd: As far as I can see the previous condition to run the production build was `tag: /v[0-9]+(\.[0-9]+)*/`. Now we have both the `branch: master` and the `tag: /v[0-9]+(\.[0-9]+)*/`. Should we only keep the tag?

beerecca: Sure

beerecca:
@spirosikmd ahh nope, think we still want that there! 
screen shot 2018-08-07 at 15 58 10

beerecca:
@spirosikmd actually we might be able to not have it, but that tags filter didn't seem to work at all!

PR #202, opened 2018-08-15
Support main navigation

beerecca:


src/components/_sidepanel.scss
```diff
@@ -1,76 +1,79 @@
...
-    }
+  &:focus {
+    outline: none;
+    border-color: $sb-color-blue-3;
+  }
 
-    &:hover {
-      border-color: $sb-color-blue-2;
-    }
+  &:focus:not(:focus-visible) {
```
spirosikmd: `focus-visible` is [not supported](https://caniuse.com/#search=%3Afocus-visible) in the browser list we mention above, right? As this code was here already, have we discussed about this?

beerecca: Yea, it's in Dine's original PR. It's a progressive enhancement. There's no way to achieve the equivalent without some js click tracking which we wanna avoid, so we're leaving it like this - it's a very small design preference, and it's more important to support being able to tab onto the handle easily.

spirosikmd: But we still keep the list of supported browsers in the comments as is, right?

beerecca: Yea cos it still works and looks good in those browsers, just looks the tinnnniiieeessst bit different

PR #204, opened 2018-08-17
fix(card): Improve card styles for react

clenemt:
:arrow_forward: **Related to**: BLOS-109
:eyeglasses: **Context**: these changes are needed for the react version of this component.
:sunrise: **After**:


src/components/_card.scss
```diff
@@ -8,20 +8,19 @@
...
 //   <div class="sb-card__header">
-//     <h1 class="sb-card__title sb-title sb-text--center">A very long and boring card title that can span over and over</h1>
+//     <h1 class="sb-title sb-text--center">A very long and boring card title that can span over and over</h1>
 //   </div>
 //   <div class="sb-card__body">
-//     <p class="sb-text sb-text--center sb-text--number sb-text--uppercase sb-text--emphasis">221 responses</p>
-//     <p class="sb-text--s sb-text--center sb-text--subtle">3 data sources</p>
+//     <button class="sb-btn sb-btn--primary">Click me</button>
 //   </div>
 // </a>
```
spirosikmd: Shouldn't this be a closing `div` tag?

src/components/_card.scss
```diff
@@ -38,25 +37,26 @@
...
-  }
-
-  &:hover,
-  &:focus {
-    outline: 0;
-    border: 1px solid $sb-color-blue-2;
-    transform: translate(0, -3px, 0);
+a.#{$ns}card:hover,
+a.#{$ns}card:focus,
+.#{$ns}card:focus-within {
```
spirosikmd: Why are we applying `hover` and `focus` on the `a` tag explicitly and `focus-within` only to the class?

src/components/_card.scss
clenemt: Noice catch 👍

src/components/_card.scss
clenemt: Because a card is only interactive (transition and border color) when it is an anchor or there is a focusable content inside

src/components/_card.scss
```diff
@@ -38,25 +37,26 @@
   border-radius: sb-px2rems($sb-radius-l);
   box-shadow: 0 2px 8px 0 rgba($sb-color-dark-2, 0.1);
   transition: all $sb-transition;
+
+  .#{$ns}title {
```
dine: why did you move away from styling `#{$ns}card__title` to just `#{$ns}title`? I think it is confusing to have styles from `#{$ns}title` inside the `_card.scss` file (and you also commented something similar here) https://github.com/usabilla/stylabilla/pull/193/files#r210531310

src/components/_card.scss
```diff
@@ -38,25 +37,26 @@
...
+a.#{$ns}card:focus,
+.#{$ns}card:focus-within {
+  outline: 0;
+  border: 1px solid $sb-color-blue-2;
+  transform: translate(0, -3px, 0);
 
-    .#{$ns}card__title {
-      color: $sb-color-blue-2;
-    }
+  .#{$ns}title {
```
dine: same as before

src/components/_card.scss
clenemt: Because it is easier when building the react component.
If I force us to add the `card__title`:
* when we pass a node to the `header="..."`, we have to add the class `card__title` every time which is very uggly

spirosikmd: But isn't it the same? As now you need to add the class `sb-title` anyway when you pass a node?

clenemt: Yea it is the same, but I feel better doing this:

```jsx
<Card header={ <h1 class="sb-title">foo</h1> } />
```

then this: 

```jsx
<Card header={ <h1 class="sb-card__title sb-title">foo</h1> } />
```

Other alternatives would be:

1- force the color override of the `sb-title` with:
```js
.#{$ns}card__header > * {
  color: $sb-color-blue-2;
}
```

which would work even if their is no `sb-title`

2- not use the `sb-title` and instead extend the `sb-card__title`:
```css
.#{$ns}card__title {
  @include sb-title();
}
```

```jsx
<Card header={ <h1 class="sb-card__title">foo</h1> } />
```

clenemt:
I fixed the stylelint plugin so we can now have regular comments in there. Please have a look again @dine @spirosikmd

lint-docs-required.js
```diff
@@ -1,44 +1,64 @@
...
+  generic: 'File is missing kss documentation',
+});
+
+module.exports = stylelint.createPlugin(ruleName, actual => (root, result) => {
+  const validOptions = stylelint.utils.validateOptions(result, ruleName, { actual });
+  if (!validOptions) return;
+
+  let commentLines = [];
+  let hasDocTexts = false;
+  let isFileParsed = true;
```
spirosikmd: As far as I can see, this never becomes `false`. What is the purpose of this variable?

lint-docs-required.js
```diff
@@ -1,44 +1,64 @@
...
+
+  root.walkComments(comment => {
+    isFileParsed = true;
+
+    const next = comment.next() && comment.next().text;
```
spirosikmd: Can we have this check `if (next !== undefined) return` here and return early? Or do we still need to add the `line` in the `commentLines` even if `next` is not undefined? If we still need it, maybe we can change the order of things?

```js
const line = comment.text;
if (line.includes('stylelint-disable')) return;

const next = comment.next() && comment.next().text;
if (next !== undefined) return;

// Then continue with the rest
const text = commentLines.join(' '); 
// ...
```

lint-docs-required.js
```diff
@@ -1,44 +1,64 @@
...
+    const line = comment.text;
+
+    // Disregard this single line if there is a `stylelint-disable` text in there
+    if (line.includes('stylelint-disable')) return;
+    commentLines.push(line);
+
+    // Process the comment once we actually have all the lines
+    if (next === undefined) {
+      const text = commentLines.join(' ');
+      const hasDocText = requiredDocTexts.some(docText => text.includes(docText));
```
spirosikmd: Can't we return directly after this line with `if (!hasDocText) return;`?

lint-docs-required.js
clenemt: Sometime a file has eslint rule disabled at top level, hence it does not get parsed

lint-docs-required.js
clenemt: We need to add the line for the comment as we are parsing each line and adding them to the array until we reach the end of the comment block. But I have removed the if indentation.

lint-docs-required.js
clenemt: Nope as we need to reset the `commentLines`. But I have move the logic around for easier read.

lint-docs-required.js
```diff
@@ -1,44 +1,66 @@
...
+  root.walkComments(comment => {
+
+    const next = comment.next() && comment.next().text;
+    const line = comment.text;
+
+    // Disregard this single line comment if there is a `stylelint-disable` text in there
+    if (!line || line.includes('stylelint-disable')) return;
+    isFileParsed = true;
+
+    console.log(comment.source.input.file);
```
spirosikmd: Do we need this `console.log` here?

src/trumps/_helpers.scss
```diff
@@ -1,3 +1,17 @@
+// Visibility Classes
+//
+// Some visibility helper classes.
+//
+// Browser support: IE9+, Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+// <div class="sb-tile sb-padding {{modifier_class}}">I'm visible</div>
```
spirosikmd: Typo: `sb-title`

PR #205, opened 2018-08-21
fix(overflow-menu): change structure adding buttons/links

dine:
Main change is wrap content on tags `button` and `a` for the role `menuitem` elements.

BREAKING CHANGE: new structure suggested to use `a` and `button` tags can bring more semantic meaning


spirosikmd:
Why is the following true?

`... new structure suggested to use less js for keyboard navigation.`

In the drop-down, when a user presses tab, they navigate outside of the drop-down, and to the next element. Why do we want a different behavior for the overflow menu? Or am I missing something?

src/components/_overflow-menu.scss
```diff
@@ -69,44 +80,48 @@
...
 
+.#{$ns}overflow-menu__header {
+  @include sb-text();
+
+  font-weight: 500;
+}
+
 .#{$ns}overflow-menu__label {
-  vertical-align: middle;
+  flex: 1;
```
spirosikmd: If I am not mistaken `flex` works in IE11+. We should update the `Browser support:` section of the docs in that case.

dine:
Updated the browser support information. This component is only going to be used in our platform so it can be IE11+ (and I think it the icons look better aligned using flex) https://support.usabilla.com/hc/en-us/articles/212566389-Which-browsers-does-Usabilla-support-

PR #206, opened 2018-08-23
feat(assets): add assets variable

clenemt:
:arrow_forward: **Related to**: BLOS-134
:eyeglasses: **Context**: we need to be able to import the fonts from `node_modules/@usabilla/stylabilla` instead of having everyone copy the assets to their own app sources.


README.md
```diff
@@ -69,9 +69,10 @@ Make sure webpack correctly packages Stylabilla by editing your `scss/css` and `
 }
 ```
 
-You can then include the css and fonts in your application with:
+Also if you want your loader to find Stylabilla fonts and images, be sure to override the assets variable before importing Stylabilla. You can then include the css and fonts in your application with:
 
 ```
+$sb-assets-path: '~@usabilla/stylabilla/src/assets/';
```
dine: 🎉

PR #212, opened 2018-09-14
[BUB-245 / BUB-249] Wizard component

MLKiiwy:
Close BUB-249

# Wizard component

- Two versions (see commits) : with/without flexbox
- Animations for opening / transitions between each steps
- aria tags for dialog role
- Follow browser support for Usabilla console : https://usabilla.atlassian.net/wiki/spaces/DEV/pages/514294083/Browser+Support

screen shot 2018-09-17 at 5 09 25 pm


dine:
I have a weird feeling about the name `wizard`. Is this full modal only going to be used for wizards? I agree that we should call things by the same name, I'm just worried that we are missing an opportunity to make something more generic here. cc @galagmon @MLKiiwy

galagmon:
@MLKiiwy the bottom bar seems to miss the shadow and the padding seems wrong though it is hard to see without the inspect from a screenshot :) 
Also the left and right padding of the whole page
and the text size and style of the status, not sure. Let me know if you want us to look at it together

MLKiiwy:
@galagmon I updated the PR, with some changes. For the Status text, I changed nothing maybe I'm not looking at the right version on Invision ? 
How do you work before to check styllabilla ? Screenshot ? A staging version running somewhere ?

galagmon:
@MLKiiwy I cannot see it in stylabilla yet I guess because it is not merged yet. can you update the screenshot?

MLKiiwy:
@galagmon  Its updated but if you want you can come see it on my computer ;)

src/components/_wizard.scss
```diff
@@ -0,0 +1,121 @@
+// Wizard
+//
+// A wizard component. ( by default not visible, you should add sb-wizard--open to open it)
+//
+// A wizard contains multiple steps and only one step is active
+//
+// Browser support: IE9+ (no transition), Safari6+, Firefox30+, Chrome30+
```
spirosikmd: We are using flex, so this should be IE11+, right?

src/components/_wizard.scss
```diff
@@ -0,0 +1,121 @@
...
+  align-self: flex-end;
+}
+
+.#{$ns}wizard__status {
+  color: $sb-color-dark-3;
+  font-feature-settings: 'lnum';
+  font-variant: small-caps;
+  font-family: $sb-primary-font-family;
+  font-size: sb-px2rems(19px);
+  line-height: sb-px2rems(40px);
```
spirosikmd: This font, in general, is the `title--s` font we have. Except the `line-height`. Why did we add `line-height` here? If it is just for the vertical alignment, can't we accomplish the same result with flex? Then we can use the `sb-title--s` class for the status.

MLKiiwy: actually only font-familly and font-size are the same as title--s. So in all the case I need a modifier for color, font-variant. I'm not using the title--s mostly because of vertical-alignement, if I want to use flex box, I need to encapsulate the status into another container, I think its a little bit too much no ?

spirosikmd: We don't need to wrap status into a container. What I meant was to use `align-items: center` in the `sb-wizard__footer` and that would align its children perfectly. Then we don't need to use `line-height`. Which then means that the `sb-wizard__status` can have the following

```scss
.#{$ns}wizard__status {
  color: $sb-color-dark-3;
  font-feature-settings: 'lnum';
  font-variant: small-caps;
}
```

and the `sb-title--s` class. What do you think?

spirosikmd: Also, we don't even need to specify the color, right? So having only the following CSS is enough for the `sb-wizard__status`:

```scss
.#{$ns}wizard__status {
  font-feature-settings: 'lnum';
  font-variant: small-caps;
}
```

MLKiiwy: Sorry you are right, I forgot this property of flexbox.

src/components/_wizard.scss
```diff
@@ -0,0 +1,116 @@
...
+//   </div>
+//   <div class="sb-wizard__footer">
+//     <span class="sb-title--s sb-wizard__status">45 status</span>
+//     <button class="sb-btn sb-btn--primary" aria-controls="wizard" aria-label="Next">Next</button>
+//   </div>
+// </div>
+//
+// Styleguide Components.wizard
+
+@keyframes fadeIn {
```
spirosikmd: Should this be a general animation that we are going to use in other cases? If yes, then maybe we should add it in the `_animations.scss` file. @galagmon what do you think?

src/components/_wizard.scss
dine: @MLKiiwy this is the first time we have this combination of font feature, variant, etc so I think it is fine here. But if we see this pattern repeating we could move it to the file where we define titles, etc.

src/components/_wizard.scss
```diff
@@ -0,0 +1,119 @@
...
+}
+
+.#{$ns}wizard__close {
+  right: sb-rem-spacing('l');
+}
+
+.#{$ns}wizard__footer {
+  width: 100%;
+  border-top: 1px solid $sb-color-light-3;
+  box-shadow: 0 -2px 10px 0 rgba(65, 71, 76, 0.1);
```
dine: Do we have this color defined? 
We should check for the other shadows we have defined and make it a variable here: https://github.com/usabilla/stylabilla/blob/933c238afde95bff4d1b3b2543539226a8687731/src/settings/_box.scss#L9

src/index.scss
```diff
@@ -67,6 +67,7 @@
 @import 'components/toast';
 @import 'components/tooltip';
 @import 'components/upload';
+@import 'components/wizard';
```
dine: I still don't like the name wizard 🙁 
Can we call everything dialog then? 😝 
https://material-ui.com/demos/dialogs/

@galagmon

spirosikmd: I like this idea a lot! And it makes sense to me! Thank you for sharing this link @dine, very useful!

MLKiiwy:
@dine @spirosikmd I change name to wizard-dialog, because its a wizard inside a dialog. 
I use now transition instead of animation, much simpler. 
I fix the padding for title, and put a variable for the box-shadow.

Please review again.

PR #213, opened 2018-09-17
feat(kss-animations): A way to see and test animations in our styleguide

MLKiiwy:
# Extension to see and test animations in our styleguide

See it here : https://static-staging.usabilla.com/stylabilla/kss-animations/docs/

Now : One thing that is missing, is the ability to see the animations of stylabilla. 
So each time we want to create an animation, we need to use the debugger and add/remove manually classes. Its also hard for designers (@galagmon @michelvdhoek) to check our work. 

This extension should solve the problem.

This extensions for our theme, provide a way to add a new control for a component that allow to see animations when you add/remove a css modifier.

For more documentation : 
https://github.com/usabilla/stylabilla/pull/213/files#diff-6a3371457528722a734f3c51d9238c13R157

Samples : 
toast
overflow_menu
modal
dropdown


docs-builder/kss-assets/kss-animations.js
```diff
@@ -0,0 +1,60 @@
+(function (window, document) {
+  'use strict';
+
+  var KssAnimations = function () {
+    var self = this;
+    document.querySelectorAll('span[data-kss-animations]').forEach(function (selectorContainer) {
+      // scope for one component
+      const section = selectorContainer.getAttribute('data-kss-animations');
+      let switchs = [];
+      document.querySelectorAll(`#kssref-${section} [data-kss-animation-switch]`).forEach(function (component) {
```
dine: We need to review some elements with `data-kss-animation-switch` because they have also the modifier markup listed below. This generates 2 switchs with the same animation. For example the Overflow Menu and the Dropdown.

docs-builder/kss-assets/kss.scss
```diff
@@ -914,6 +914,13 @@ ul.kss-tabs {
   text-decoration: underline;
 }
 
+// kss-animations
+
+.kss-animations-selectors {
+  display: inline-flex;
+  flex-direction: row;
+}
+
```
dine: Can we add a little bit more styling?

```
.kss-animations-selectors__label {
  margin-left: 10px;
}

.kss-animations-selectors__select {
  font-size: 13px;
}  
````

File: docs-builder/kss-assets/kss-animations.js [hunk 3]
```diff
@@ -0,0 +1,60 @@
...
+
+      switchs.forEach(element => {
+        const label = document.createElement('label');
+        label.innerHTML = 'Animations : ';
+
+        const select = document.createElement('select');
+        const option = document.createElement('option');
+        option.innerHTML = 'No modifiers';
+        option.setAttribute('value', '');
+        select.appendChild(option);
```
dine: `select.classList.add('kss-animations-selectors__select');`

docs-builder/kss-assets/kss-animations.js
```diff
@@ -0,0 +1,60 @@
...
+
+        element.classes.forEach(function(className) {
+          const option = document.createElement('option');
+          option.innerHTML = className;
+          option.setAttribute('value', className);
+          select.appendChild(option);
+        });
+
+        label.appendChild(select);
```
dine: `select.classList.add('kss-animations-selectors__label');`

docs-builder/kss-assets/kss.scss
MLKiiwy: Done

docs-builder/kss-assets/kss-animations.js
MLKiiwy: Done

galagmon:
I can't see the selector for the animation in Stylabilla. do I need a new link?

docs-builder/index.hbs
```diff
@@ -249,6 +251,7 @@
     bodyClass: 'kss-markup-mode',
     detailsClass: 'kss-markup'
   });
+  var kssAnimations = new KssAnimations();
```
spirosikmd: Do we need to set it to a variable? Isn't `new KssAnimations()` enough?

docs-builder/kss-assets/kss.js
```diff
@@ -5,6 +5,7 @@ var prettify = require('./prettify.js');
 var fullscreen = require('./kss-fullscreen.js');
 var guides = require('./kss-guides.js');
 var markup = require('./kss-markup.js');
+var animations = require('./kss-animations.js');
```
spirosikmd: Also this, it is enough to just `require('./kss-animations.js');`, right?

docs-builder/kss-assets/kss-animations.js
```diff
@@ -0,0 +1,97 @@
+(function (window, document) {
```
spirosikmd: Aren't these JS files transpiled with the KSS project? If yes, should we use ES6 syntax?

docs-builder/kss-assets/kss-animations.js
```diff
@@ -0,0 +1,97 @@
...
+  'use strict';
+
+  const createGlobalSwitcher = function(section, selectorContainer) {
+    let switchs = [];
```
spirosikmd: We can write this a bit differently to avoid filling the array of switches with objects. For example

```js
const switches = document.querySelectorAll(`#kssref-${section} [data-kss-animation-switch]`);

if (switchs.length === 0) {
  return;
}

switchs.forEach(element => {
  // element is the component in here
  const classes = element.getAttribute('data-kss-animation-switch').split(',');

  // The rest of the implementation here ...
});
```

CONTRIBUTING.md
```diff
@@ -153,3 +153,31 @@ This includes:
...
+
+Example:
+
+```html
+<button class="sb-btn {{modifier}}" data-kss-animation-cb="sb-dropdown--open"></button>
+```
+
+See also "Dropdown" or "Overflow Menu" component.
+
+## Animation on a global example
```
spirosikmd: Is this supposed to use `###` instead of `##`?

CONTRIBUTING.md
```diff
@@ -153,3 +153,31 @@ This includes:
...
+Put a property "data-kss-animation-cb" on the element you want to animate in the markup.
+The property should contain the modifier.
+
+Example:
+
+```html
+<button class="sb-btn {{modifier}}" data-kss-animation-cb="sb-dropdown--open"></button>
+```
+
+See also "Dropdown" or "Overflow Menu" component.
```
spirosikmd: Should we add links to `Dropdown` and `OverflowMenu` components?

CONTRIBUTING.md
```diff
@@ -153,3 +153,31 @@ This includes:
...
+Put a property "data-kss-animation-switch" on the element you want to animate in the markup.
+The property should contain all the classes can apply for animate the element.
+
+Example:
+
+```html
+<button class="sb-btn" data-kss-animation-switch="sb-btn--anim1,sb-btn--anim2"></button>
+```
+
+See also the "Modal" component.
```
spirosikmd: Should we add link to `Modal` component?

MLKiiwy:
@spirosikmd I rewrite it to es6, you're right its in the transpiler !

For the choice between selector and checkbox.

In case the animation is on a modifier, because we display each of the modifiers, its better to use a checkbox. Simple on/off.

For the selector, its more for cases like the Modal, where we can have an animation on multiple css classes (one for open / one for close) and we don't consider these classes as modifiers.

For the SidePanel, its more complex, right now the plugin can't handle it properly because to animate it you need to put a class on multiple elements in the same time. Maybe I should remove this one ?

docs-builder/kss-assets/kss-animations.js
```diff
@@ -0,0 +1,98 @@
...
+            component.classList.remove(className);
+          }
+        });
+      }
+    });
+};
+
+const setup = () => {
+  document
+    .querySelectorAll('span[data-kss-animations]')
```
spirosikmd: Why are we limiting ourselves to `span` elements?

docs-builder/kss-assets/kss-animations.js
```diff
@@ -0,0 +1,98 @@
...
+    });
+  });
+};
+
+const createCheckboxForModifiers = section => {
+  document
+    .querySelectorAll(`#kssref-${section} [data-kss-animation-cb]`)
+    .forEach(component => {
+      const className = component.getAttribute('data-kss-animation-cb');
+      if (!className || className === '') {
```
spirosikmd: `!className` also covers `className === ''`, so the second statement is redundant!

PR #214, opened 2018-09-19
feat(table): add table component

dine:
## Updated screenshot:

localhost_8080_item-components-table html ipad 1

## Latest changes:

- [x]  Remove checkbox from table header
- [x]  Align checkboxes
- [x]  Apply default alignment `th` to component (remove from elements)

## Pending
- [x]  Check with @galagmon or @michelvdhoek the dot alignment

https://static-staging.usabilla.com/stylabilla/grid/docs/item-components-table.html


src/components/_icons.scss
```diff
@@ -200,3 +200,32 @@
...
+//
+// Markup:
+// <div class="sb-icon__status {{modifier_class}}"></div>
+//
+// .sb-icon__status--success - Applies a successful color to your icon.
+// .sb-icon__status--error   - Applies a error color to your icon.
+//
+// Styleguide Components.icons.status
+
+.#{$ns}icon__status {
```
spirosikmd: Why don't we use SVG for these icons as well? Then we can also reuse the `sb-icon--success` and `sb-icon--error` classes.

src/components/_table.scss
```diff
@@ -0,0 +1,218 @@
...
+// </div>
+//
+// Styleguide Components.table
+
+.#{$ns}table {
+  width: 100%;
+  min-width: $sb-breakpoint-m;
+}
+
+.#{$ns}table__container {
```
spirosikmd: Why are we defining this here? The table should not know where it is rendered, right? The parent, which can be any parent can add the overflow to make the table use scrolling.

src/components/_table.scss
```diff
@@ -0,0 +1,218 @@
...
+//         </tr>
+//      </tbody>
+//   </table>
+// </div>
+//
+// Styleguide Components.table
+
+.#{$ns}table {
+  width: 100%;
+  min-width: $sb-breakpoint-m;
```
spirosikmd: Should we use the [`table* display`](https://developer.mozilla.org/en-US/docs/Web/CSS/display) properties?

src/components/_table.scss
```diff
@@ -0,0 +1,218 @@
...
+}
+
+.#{$ns}table__container {
+  overflow-x: auto;
+}
+
+.#{$ns}table__tr,
+.#{$ns}table__sort {
+  cursor: pointer;
```
spirosikmd: Why are we using the pointer by default? A table can be used to just show data. In that case, no action would be required, right?

src/components/_table.scss
```diff
@@ -0,0 +1,218 @@
...
+  padding: sb-rem-spacing('l') sb-rem-spacing('s');
+  vertical-align: middle;
+}
+
+.#{$ns}table__icon,
+.#{$ns}table__checkbox {
+  visibility: hidden;
+}
+
+.#{$ns}table__hh[aria-sort='ascending'] {
```
spirosikmd: What about descending?

galagmon:
@dine can you update the screenshot base on the comments we already went through so I can review? :)

src/components/_table.scss
dine: Hmm, not sure I understand 🤔
What do you mean here? The default browser display for `<table>` tag is `display: table` already... Unless we would use this class with different markup, which I think we should avoid for now, and we don't see the need yet...

src/components/_table.scss
dine: This changed completely, sorry, you reviewed a work-in-progress 🙂

src/components/_table.scss
```diff
@@ -0,0 +1,248 @@
...
+//
+// Browser support: IE9+, Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+// <div class="sb-overflow-x">
+//   <table class="sb-table {{modifier_class}}">
+//      <thead class="sb-table__thead sb-text--smallcaps">
+//         <tr class="sb-table__tr">
+//            <th class="sb-table__hh sb-table__action sb-text--center" scope="col">
+//              <div class="sb-table__checkbox sb-checkbox">
```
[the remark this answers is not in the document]
dine: Some of them are back now :)

src/trumps/_helpers.scss
```diff
@@ -19,3 +19,7 @@
 .#{$ns}completely-hidden {
   display: none;
 }
+
+.#{$ns}overflow-x {
```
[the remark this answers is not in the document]
dine: Like discussed IRL, this is part of the responsiveness of the component. We can use it as a helper, in the component, or even using inline styles, but in my opinion it should be part of it.

src/components/_wizard-dialog.scss
```diff
@@ -30,7 +30,7 @@
 //     </div>
 //   </div>
 //   <div class="sb-wizard-dialog__footer">
-//     <span class="sb-title--s sb-wizard-dialog__status">45 status</span>
+//     <span class="sb-title--s sb-text--smallcaps">45 status</span>
```
MLKiiwy: 👍

src/components/_table.scss
spirosikmd: You are right! We don't need to use these properties!

src/elements/_tables.scss
```diff
@@ -0,0 +1,4 @@
+th {
```
spirosikmd: Do we need to add this CSS here? Can't it be part of components table CSS?

kagnamazian:
Good job @dine !! 
Yes, there is something weird on the checkbox that makes the thing no aligned.
basically, the div 'sb-checkbox' has the dimension: 16*20 but the input checkbox itself is 16*16
It doesn't come from any padding but I do see the difference.

galagmon:
As mentioned the alignment of the check box :)
I would also love to check the sorting and the acceptability. If this should be merged before Thursday or even Friday please ask @michelvdhoek to check it

src/elements/_tables.scss
dine: I don't mind, but this means more CSS lines being generated, or more markup required. Since we don't know the future, I'll move to the component

PR #215, opened 2018-09-26
chore(build): deploy to staging on pull request

dine:
Temporary until we have a test/dev bucket/environment, this way designers can view and interact with components before we merge the PR.


.circleci/config.yml
```diff
@@ -24,6 +24,18 @@ jobs:
             yarn test:accessibility
       - run: DEBUG=condition yarn semantic-release || true
 
+  deploy_staging_branch:
+    docker:
+      - image: circleci/node:8
```
renatomefi: `node:8-alpine` is smaller!

.circleci/config.yml
```diff
@@ -24,6 +24,18 @@ jobs:
...
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - v1-dependency-cache-{{ checksum "yarn.lock" }}
+            - v1-dependency-cache-
```
renatomefi: What's this empty suffix?

spirosikmd:
And the deployed code can be accessed in:

https://static-staging.usabilla.com/stylabilla/test-staging/docs/ 

This is cool 🎉

.circleci/config.yml
spirosikmd: You could use the `circleci/node:carbon` as well, which points to the latest stable LTS node release. It would nice in this case to change `.nvmrc` to `lts/carbon` for the same reason.

.circleci/config.yml
```diff
@@ -24,6 +24,18 @@ jobs:
             yarn test:accessibility
       - run: DEBUG=condition yarn semantic-release || true
 
+  deploy_staging_branch:
```
spirosikmd: We could name this `deploy_branch_to_staging` to make it more explicit. Now it reads as if we are deploying the staging branch.

.circleci/config.yml
```diff
@@ -54,6 +66,7 @@ workflows:
   pull_request:
     jobs:
       - test_and_npm_release
+      - deploy_staging_branch
```
spirosikmd: Should add a `requires` here for `test_and_npm_release`?

.circleci/config.yml
dine: This is used if the first checksum fails
https://circleci.com/docs/2.0/caching/#example-caching-configuration

PR #217, opened 2018-10-04
feat(date-picker): implement date picker component

spirosikmd:
What is covered so far:
* The basic calendar with months, weeks, and days. The HTML is taken
from the output HTML of `react-day-picker`.
* Actions at the bottom for clear and apply.
* Navigation bar at the top to navigate between months and years.
* The button that shows the selected dates and handles the open/close
interaction.

Link to component demo: https://static-staging.usabilla.com/stylabilla/date-picker/docs/section-components.html#kssref-components-date-picker

### Visual Changes

screencast 2018-10-04 12-55-38

#### Disabled

screen shot 2018-10-04 at 17 19 53


src/settings/_box.scss
```diff
@@ -9,7 +9,7 @@ $sb-radius-infinite: 9999em;
 $sb-box-shadow: 0 1px 2px transparentize($sb-color-black, 0.85);
 $sb-box-shadow-subtle: 0 1px 0 0 $sb-color-light-3, 0 1px 2px transparentize($sb-color-black, 0.85);
 $sb-box-shadow-light: 0 2px 4px transparentize($sb-color-black, 0.8), 0 1px 0 0 transparentize($sb-color-light-1, 0.8);
-$sb-box-shadow-heavy: 0 1px 0 0 $sb-color-light-3, 0 2px 4px transparentize($sb-color-black, 0.5);
+$sb-box-shadow-heavy: 0 2px 8px 0 rgba(65, 71, 76, 0.1);
```
MLKiiwy: it's not affecting other component ? And we should avoid const here and use variables for colors

MLKiiwy:
Adding url for @galagmon : https://static-staging.usabilla.com/stylabilla/date-picker/docs/section-components.html#kssref-components-date-picker

src/components/_date-picker.scss
```diff
@@ -0,0 +1,570 @@
...
+//
+// A disabled date picker.
+//
+// Browser support: IE11+, Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+// <div class="sb-date-picker">
+//   <button class="sb-date-picker__button sb-date-picker__button--disabled">
+//     <div class="sb-flex-inline sb-justify-content-center sb-align-items-baseline sb-padding-left-xs sb-padding-right-xs">
+//       <div class="sb-text--s sb-date-picker__label">13/09/2018</div>
```
MLKiiwy: there is no way to use input instead of div ? because that mean we must use datepicker in all cases and not entering the date manually.

galagmon:
@spirosikmd 
- The arrows should be a perfect square. currently, there is too much right and left padding.  
- There is an opacity transition missing on the hover 
- can we add the open and close interaction so I can see that? 
- I added a focused calendar design to the style guide 
image

src/components/_date-picker.scss
dine: I think the markup is generated by http://react-day-picker.js.org/

spirosikmd: This specific markup was our choice. We are free to use any markup we want for the date picker button. Initially, we were using inputs, but then we decided to try it out without inputs in order to reduce complexity. Also, the date picker will have another drop down with shortcuts (select last week, last month, last six months etc.). However, judging from the result, I am very confident that using two inputs will improve the user experience. So we'll update!

src/settings/_box.scss
spirosikmd: The date picker is the only place this shadow is used. I agree we should use a color variable!

MLKiiwy:
@spirosikmd not completely true, we defined an implementation and behavior in react-stylabilla but animation are defined here : https://github.com/usabilla/stylabilla/pull/217/files#diff-98873acd215d8f248f5042e7bfc0e095R242 
So yes, @galagmon should be able to see the animation in action before merging, otherwise if the animation is not valid we need to modify this project again.
That's why I have done : https://github.com/usabilla/stylabilla/pull/217/files#diff-98873acd215d8f248f5042e7bfc0e095R242

spirosikmd:
True @MLKiiwy! Let's try and get #213 merged!

src/components/_date-picker.scss
```diff
@@ -0,0 +1,593 @@
...
+// JavaScript is required for:
+// - Opening and closing the date picker.
+// - The whole calendar interaction: selecting dates, navigating to months, interacting with the actions.
+// - Updating the dates on the button with the selected dates.
+//
+// Markup:
+// <div class="sb-date-picker">
+//   <button class="sb-date-picker__button">
+//     <div class="sb-flex-inline sb-justify-content-center sb-align-items-baseline sb-padding-left-xs sb-padding-right-xs">
+//       <div class="sb-text--s sb-date-picker__label">13/09/2018</div>
```
dine: Missing using the `<inputs>`, like mentioned here: https://github.com/usabilla/stylabilla/pull/217/files#r222791350

src/components/_button.scss
```diff
@@ -26,7 +26,7 @@
 //
 // Styleguide Components.button
 
-$sb-btn-padding: sb-rem-spacing('s') sb-rem-spacing('m');
+$sb-btn-padding: sb-rem-spacing('s');
 $sb-btn-padding-s: sb-px2rems(7px);
 $sb-btn-padding-l: sb-rem-spacing('s') sb-rem-spacing('l');
```
kagnamazian: is this one should also be changed? it does look weird.

src/components/_date-picker.scss
spirosikmd: We are going to implement that in a subsequent PR. The first version of the date picker will not use inputs!

src/components/_button.scss
spirosikmd: Have a look [this commit](https://github.com/usabilla/stylabilla/pull/217/commits/02099d689f88cfcd2f0f3f870c3eb63c9644dc0a), where we explain the reasons behind this change.

PR #220, opened 2018-10-09
feat(date-picker): add select month and year options

spirosikmd:
[Demo with the changes](https://static-staging.usabilla.com/stylabilla/select-month-year/docs/section-components.html#kssref-components-date-picker)

### Visual Changes

screencast 2018-10-09 20-16-16


src/components/_date-picker.scss
```diff
@@ -48,9 +50,37 @@
 //           </div>
 //           <div class="sb-date-picker__months">
 //             <div class="sb-date-picker__month" role="grid">
-//               <div class="sb-date-picker__caption" role="heading">
-//                 <div>September 2018</div>
-//               </div>
+//               <form class="sb-date-picker__caption" role="heading">
+//                 <div class="sb-date-picker__caption-select-wrapper sb-margin-right-xs">
```
dine: Right now the month and year look too far away, but maybe I'm wrong. I think it would look better using `sb-margin-right-xxs`

src/components/_date-picker.scss
```diff
@@ -361,6 +392,7 @@
 .#{$ns}date-picker__prev-month {
   position: absolute;
```
dine: They will always move together, so I think we should group `position: absolute` and `top: -1px` and set for `__next-month` and `__next-month` at the same time.

src/components/_date-picker.scss
spirosikmd: It looks different due to the white space. This is how it looks with `xxs` and whitespace:

screenshot 2018-10-10 at 11 46 12

and this is how it looks with `xs` and without whitespace:

screenshot 2018-10-10 at 11 45 35

They look the same!

src/components/_date-picker.scss
```diff
@@ -373,9 +406,43 @@
...
+  }
+}
+
+.#{$ns}date-picker__caption-select {
+  border: none;
+  border-radius: 0;
+  appearance: none;
+  position: absolute;
+  outline: none;
+  z-index: 2;
```
kagnamazian: don't we have a pre-z-index we could re-use? I think we have dropdown z-index property

spirosikmd: These values are local to the date picker calendar. The select has a z-index of 2 and the wrapper of 1. This is to have the select element on top of the text so when we click, we get the select element and select months or years.

PR #221, opened 2018-10-12
fix(wizard-dialog): content overflow

MLKiiwy:
Reviewers: @usabilla/fe-chapter @galagmon @michelvdhoek 

- [x] Add a new example with a content that need a scrollbar and use the modifier : --srcollable
- [x] by default content is not scrollable and should grow as maximum

Test it here : https://static-staging.usabilla.com/stylabilla/BUB-246-fix-wizard-dialog-content-container/docs/index.html


docs-builder/kss-assets/kss.scss
```diff
@@ -955,15 +955,18 @@ ul.kss-tabs {
 .kss-modifier__example .sb-wizard-dialog {
   position: relative;
   overflow: auto;
+  height: 350px;
```
kagnamazian: we do need to fix this height?

src/components/_wizard-dialog.scss
```diff
@@ -92,6 +103,7 @@
   flex-direction: row;
   justify-content: space-between;
   align-items: center;
+  min-height: 68px;
```
dine: This sounds like a magic number. We could create a variable like `$sb-wizard-footer` or something like that here: https://github.com/usabilla/stylabilla/blob/master/src/settings/_box.scss#L24.

src/components/_wizard-dialog.scss
```diff
@@ -101,9 +113,86 @@
 
 .#{$ns}wizard-dialog__step {
   opacity: 0;
+  flex-grow: 1;
+  flex-shrink: 1;
+  flex-basis: auto;
```
dine: Same about the shorthand

src/components/_wizard-dialog.scss
```diff
@@ -65,9 +67,18 @@
 }
 
 .#{$ns}wizard-dialog__content {
+  flex-grow: 1;
+  flex-shrink: 1;
+  flex-basis: auto;
+  display: flex;
+  overflow-y: hidden;
```
dine: Instead of having the modifier class `.#{$ns}wizard-dialog__content--scrollable`, what are the implications if we use here `overflow-y: auto`?

galagmon:
@MLKiiwy this looks really great! why do we need the centered button in the first one?
Also, will we also have the possibility to scroll within the page but not the whole page? 
(like in the sharing screen https://projects.invisionapp.com/d/main#/console/15239297/316840026/preview)

src/components/_wizard-dialog.scss
MLKiiwy: if you put "auto" you are going to have scrollbar, that's the thing I want to avoid in the generic case. Most of the time if I check the prototypes on InVision, we are going to put a scrollbar on an a sub-element, so not on the entire content of the wizard.

MLKiiwy:
@galagmon the centered button is an example here to indicate how to center an element inside this component. I have seen that in many cases we are going to center align component in the wizard dialog. 
For the scolling, actually in the actual prototype most of the time the scrolling will be on a child component of the wizard, that's why by default no scrolling is apply on the content. The scrolling should be implemented on the child component.

src/components/_wizard-dialog.scss
dine: @MLKiiwy which browser are you seeing the scrollbar? Just checked in Chrome and it worked fine. For the first example the content is not big enough, so `overflow: auto` will only add scrollbars if needed. To always adds a scrollbar is `overflow: scroll` (default value is `visible`) https://developer.mozilla.org/en-US/docs/Web/CSS/overflow

src/settings/_box.scss
```diff
@@ -27,3 +27,6 @@ $sb-main-navigation-width: 75px;
 $sb-main-navigation-overflow-menu-left: $sb-main-navigation-width - 11px;
 $sb-container-margin: $sb-sidepanel-width + $sb-main-navigation-width;
 $sb-overflow-menu-min-width: 150px;
+
+// WizardDialog
+$sb-wizard-dialog-footer-height: 68px;
```
spirosikmd: Why are we adding this variable here? Is it only used in the `_wizard-dialog.scss` file?

src/components/_wizard-dialog.scss
spirosikmd: I agree with @dine that `overflow: auto` should be enough.

Also, I don't see the scrollbar when scrolling. Is it me or is this normal? Shouldn't the scrollbar appear?

screencast 2018-10-15 12-02-21

src/settings/_box.scss
dine: I suggested here, following the idea of ITCSS (we always define variables in the settings). But you are right, if we don't need to use it anywhere else can be moved to `_wizard-dialog.scss`

MLKiiwy:
@spirosikmd @dine you are both right, I switch to overflow: auto and remove the modifier, and the second example. That mean here, the container allow scrolling only when its needed by the subcomponents. So the config is set on the content not on the container, its simpler.

PR #222, opened 2018-10-16
fix(table): add sb-table--scrollable-content modifier and example

MLKiiwy:
Reviewers: @dine @MLKiiwy @spirosikmd @kagnamazian 

# Table with scrollable content

See result : https://static-staging.usabilla.com/stylabilla/BUB-246-scrollable-table-grid-content/docs/section-components.html#kssref-components-table-scrollable

I implement a new modifier for setting a scrollable content.
I also did some change in combinaison with other modifier.

Todo : check on other browsers

Two issue identified but not linked to this PR : 
- on IE / Edge smallcaps is not applied to THEAD
- on safari there is a center align on form name column ...
@spirosikmd @kagnamazian I open another PR to adress these one ?


galagmon:
Is it the browser scrollbar? can we customize it?

galagmon:
@MLKiiwy it changed the alignment of the right column 
image

MLKiiwy:
@galagmon yes, right now its always the browser scrollbar.
We can customize the scrollbars only on Safari/Chrome, maybe on other browser but in that case we need a JS plugin I think.

MLKiiwy:
I have completely updated the code, now everything should be aligned.

The only thing compare to a standard table : 
- not responsive at all (fixed width)
- cannot put an horizontal scrollbar now
- need to specified column width

src/components/_table.scss
```diff
@@ -263,3 +263,514 @@
...
+// - your table needs to be in a parent with a fixed size
+// - each column (th and td elements) need to have a min-width specified (In this example we define it into a specific style html block)
+// - there is also a specific class for tbody (sb-table__tbody)
+//
+// Browser support: IE9+, Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+// <style>
+//   #kssref-components-table-scrollable td,
+//   #kssref-components-table-scrollable th {
```
dine: I'm not sure here is the best place to add this code, but I understand you want to make it public the sizes you are adding manually.

src/components/_table.scss
```diff
@@ -263,3 +263,514 @@
...
+//              </svg>
+//              Active
+//            </td>
+//         </tr>
+//      </tbody>
+//   </table>
+// </div>
+//
+// .sb-table--checkbox-visible - Shows all elements on checkbox column.
+// .sb-table--compact.sb-table--checkbox-visible - Applies small padding to the cells of your table.
```
dine: I think we don't need to add all the variations twice, this block should be only to show the scrollable version. Better if we could add `.#{$ns}table--scrollable-content` to the first block.

src/components/_table.scss
```diff
@@ -263,3 +263,514 @@
...
+// .sb-table--comfortable.sb-table--checkbox-visible - Applies big padding to the cells of your table.
+//
+// Styleguide Components.table.scrollable
+
+.#{$ns}table-container {
```
dine: We had discussed before about adding a container as part of the component https://github.com/usabilla/stylabilla/pull/214#discussion_r219079295

src/components/_table.scss
```diff
@@ -263,3 +263,514 @@
...
+
+  .#{$ns}table__tbody {
+    width: 100%;
+    height: 100%;
+    overflow-y: auto;
+    overflow-x: hidden;
+    display: block;
+  }
+
+  .#{$ns}table__thead {
```
dine: did you try `display: flex` to the thead, like they do in Dropbox? It might be like the `table-layout: fixed` thing, but then we can use percentages and make it more responsive

src/components/_table.scss
dine: I also don't like the idea of using fixed sizes 😕

src/components/_table.scss
dine: But then we'd have to add the `position:fixed` to the `thead>tr` and always calculate the width based on the page elements where the table is. If we could only use `position:sticky` 🤔

MLKiiwy:
Actually even with div and flexbox we face the same issue with alignment due to scrollbar.
So right now I see only two solutions : 

- using my first version with positioning thead absolute, you have to specify the width of each column but the width can be a percent. Issue : when there is a scrollbar nothing is align, so we need a specific code in JS launch when application load that check if scrollbar are display + size (https://gist.github.com/sttk/164f82cef5d9c5ab218211fff6d68e5d) and after we apply a move on the header.

- using my second version, everything is always align but no responsive column ( everything have a fixed size in px)

galagmon:
@MLKiiwy hey but now you cannot read the input in the right column and the right header.
Also is it still flexible? will be able to add a different amount of columns?
image

MLKiiwy:
@galagmon yep that's exactly the issue with this solution, everything is align but column are not flexible, = fixed width for the table. 
I think my previous solution was better, but require adding some JS in react to fix the alignment

kagnamazian:
@MLKiiwy is this PR ready to review ?

MLKiiwy:
I close this one, and open a new one.

PR #224, opened 2018-10-18
feat(table-scrollable): add sb-table--scrollable-content modifier and example

MLKiiwy:
Reviewers: @dine @MLKiiwy @spirosikmd @kagnamazian 

Add support for scrolling on table component. 
Using a modifier on table component.

See result : https://static-staging.usabilla.com/stylabilla/BUB-246-scrollable-table-grid-content/docs/section-components.html#kssref-components-table-scrollable

In the example the thead column are not aligned in some browsers (when scrollbar are visible and integrated to content). The alignment will be done in the react-styllabilla component.


src/components/_table.scss
```diff
@@ -263,3 +263,543 @@
...
+// .sb-table--comfortable.sb-table--checkbox-visible - Applies big padding to the cells of your table.
+//
+// Styleguide Components.table.scrollable
+
+$sb-table-header-height: 50px;
+$sb-table-header-height-compact: 40px;
+$sb-table-header-height-comfortable: 63px;
+$sb-table-checkbox-width: 40px;
+
+.#{$ns}table-container {
```
dine: If we are ok with `sb-table-container` we might not need the class `sb-overflow-x-auto`. If we always have a wrapper on the table, we might as well use the outside div to apply the modifier class. Like the renaming suggestions I made above. What do you think?

src/components/_table.scss
```diff
@@ -263,3 +263,543 @@
...
+//   #kssref-components-table-scrollable th:nth-child(1) {
+//     width: 40px;
+//   }
+//   #kssref-components-table-scrollable td:nth-child(2),
+//   #kssref-components-table-scrollable th:nth-child(2) {
+//     width: 50%;
+//   }
+// }
+// </style>
+// <div class="sb-table-container sb-overflow-x-auto">
```
dine:
```suggestion
// <div class="sb-table sb-table--scrollable-content {{modifier_class}}">
```
Related to comment below.

src/components/_table.scss
```diff
@@ -263,3 +263,543 @@
...
+//   <table class="sb-table sb-table--scrollable-content {{modifier_class}}" role="grid">
```
dine:
```suggestion
//   <table class="sb-table__table" role="grid">
```
Related to comment below.

MLKiiwy:
All comments adressed, no more "container" instead sb-table is the first component and there is a sb-table__table element

PR #225, opened 2018-10-18
docs(icons): adds main navigation icons

dine:
Add new icons.
icons


spirosikmd:
Why are we adding these icons to this project? Where are they meant to be used?

PR #226, opened 2018-10-30
feat(badges): implement badges component

realdreamer:
reviewers: @usabilla/fe-chapter


spirosikmd:
For reference, this is the generated docs link for this PR:

https://static-staging.usabilla.com/stylabilla/BLUP-86/badge-component/docs/section-components.html#kssref-components-badge

galagmon:
A few comments:
Naming:
We shouldn't use naming as 'success' 'warning' and so one, we should be able to apply any usabilla color to it. The use case is based on the average score of the NPS and that includes multiple usabilla colors.
Also if we what to have the same 'badges' used for rating we should be able to apply color on the text and border but not the background. 
image

Sizes: 
We should probably introduce the usabilla sizes here.
The Default one we have now should be actually L
The small one is in the tooltip

There should be an XL one in case of one-day results 
image

As we just discuss if we need more refinement and better planning for this component to be flexible let's do that!

Also, don't forget to add Marco as a reviewer as well!

realdreamer:
> So basically the colors there are just an example and you can apply any color to the background, the border and the text right? 
Yes, those are just an example. You can add any background border text classes to customize it 

> Is the text in the badge also using the stylabilla text classes?
Yes.  

> image
> the border color doesn't match the text on the right, what does this text mean? 
Sorry. Forgot to update the text. I'll fix it

galagmon:
@realdreamer Great work! 🎉

src/components/_badge.scss
```diff
@@ -0,0 +1,113 @@
...
+  min-height: $extra-large-min-height;
+  min-width: $extra-large-min-width;
+  padding: 0 sb-rem-spacing('xs');
+}
+
+.#{$ns}badge--rounded {
+  border-radius: $sb-radius-infinite;
+}
+
+// Background classes
```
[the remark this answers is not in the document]
realdreamer: Agreed. when we just scroll its confusing. But, Badge variations with background classes looks little big, so instead will go with "Badge with background classes". what do you say?

src/components/_badge.scss
```diff
@@ -0,0 +1,112 @@
...
+//
+// Markup:
+//  <span class="sb-badge {{modifier_class}}">
+//    -50
+//  </span>
+//
+// .sb-badge--s - Small size badge
+// .sb-badge--l - Large size badge
+// .sb-badge--xl - Extra large size badge
+// .sb-badge--rounded - Rounded badge
```
kagnamazian: I guess in the future we might need to have `sb-badge--rounded--l` but for now it's ok

src/components/_badge.scss
```diff
@@ -0,0 +1,112 @@
...
+//
+// .sb-color-blue-1  - Text blue 1
+// .sb-color-green-1 - Text green 1
+// .sb-color-yellow-1 - Text yellow 1
+//
+// Styleguide Components.badge.color
+
+// Badge with border classes
+//
+// Browser support: IE3+, Safari1+, Firefox1+, Chrome10+
```
kagnamazian: it does feel a little weird that Badge support from IE9 but badge with border classes support IE3. I suppose you mean only the border classes. So I would propose if you put it under Badge categories, The browser support doesn't change

src/tools/_shared.scss
```diff
@@ -31,6 +31,12 @@
   font-size: sb-px2rems(13px);
 }
 
+@mixin sb-text-xl {
```
kagnamazian: I think you meant `sb-text-xl()`

src/components/_badge.scss
realdreamer: if apply both `sb-badge--l and sb-badge--rounded` it works

kagnamazian: ok good !

PR #230, opened 2018-11-21
feat(milo-modifier): add sb-text--milo modifier

MLKiiwy:

Add a new modifier to force using milo.


src/trumps/_text.scss
```diff
@@ -20,6 +20,9 @@
 // .sb-text--uppercase - Used for uppercase text
 // .sb-text--lowercase - Used for lowercase text
 // .sb-text--number    - Used to reset numbers to their original alignment
+// .sb-text--number    - Used to reset numbers to their original alignment
+// .sb-text--smallcaps - Used to display text in small caps
+// .sb-text--milo      - Force Milo font
```
kagnamazian: is it something than Gal ask ?

MLKiiwy: it's ask by @Maruceau

MLKiiwy: Ok @kagnamazian it's good for now, we can add this modifier. I think later we are going to do a refactoring with less modifier when @Maruceau will be ready

src/trumps/_text.scss
```diff
@@ -20,6 +20,9 @@
 // .sb-text--uppercase - Used for uppercase text
 // .sb-text--lowercase - Used for lowercase text
 // .sb-text--number    - Used to reset numbers to their original alignment
+// .sb-text--number    - Used to reset numbers to their original alignment
```
dine: `.sb-text--number` is duplicated :)

PR #234, opened 2018-11-29
feat(table-row-hover): disable table row hover highlight

realdreamer:
Review: @usabilla/fe-chapter 

## Changelog:

- [x] disable border highlighting when hover table row


kagnamazian:
https://static-staging.usabilla.com/stylabilla/disable-table-row-hover/docs/section-components.html#kssref-components-table

Hey I don't really see the difference. Tomorrow can you show me the changed ? Otherwise code wise it look fine

PR #238, opened 2018-12-05
feat(chip): implement chip component

realdreamer:
Reviewers: @MLKiiwy @kagnamazian @realdreamer @galagmon @Maruceau @michelvdhoek

Preview: [https://static-staging.usabilla.com/stylabilla/pr/blup-83-filter-tag/docs/section-components.html](https://static-staging.usabilla.com/stylabilla/pr/blup-83-filter-tag/docs/section-components.html)


src/settings/_colors.scss
```diff
@@ -140,6 +140,9 @@ $sb-color-mood-background: #ffd442;
 $sb-color-mood-heart-eyes: #f4606e;
 $sb-color-mood-lines: #594504;
 
+// Error color
+$sb-color-error: #feeef0;
```
kagnamazian: in the input component and many other, we are using a red for the error. Are we changing the error colour for them as well? We should stay consistent.

src/components/_filter-tags.scss
```diff
@@ -0,0 +1,246 @@
+// Filter Tag
```
kagnamazian: I think this name is confusing because we also have "tags" and actually these `filter tags` might not be just for filter only. I guess we should stick to query builder component. Or do you have an other idea ?

realdreamer:
@Maruceau 
There should be 11 pixel spacing between rows of tags, not there is no spacing. Or is this part of the query builder container and not the tag? 

-- spacing adjustments between the tags actually will do the in the query-builder-container

In the "Filter tag with delete" example, Platform is truncated and the string of platforms after "Fi...". What is the rule when to truncate? I think truncating Platform into Platfo... doesn't make a lot of sense.

There seems to be quite a bit of space between the condition and the range 1 - 5, why is that bigger than the list of platforms?

-- regarding the spacing we can't exactly set the width of the text based on the each tag text, so I discussed with Gal and set it as the min-width: 30px. so based on other tag content width its adjust. so if you want to set the different min-width for the different content also we can set it. will discuss and change it accordingly.

realdreamer:
@Maruceau 
There should be 11 pixel spacing between rows of tags, not there is no spacing. Or is this part of the query builder container and not the tag? 

-- spacing adjustments between the tags actually will do it in the query-builder-container

In the "Filter tag with delete" example, Platform is truncated and the string of platforms after "Fi...". What is the rule when to truncate? I think truncating Platform into Platfo... doesn't make a lot of sense.

There seems to be quite a bit of space between the condition and the range 1 - 5, why is that bigger than the list of platforms?

-- regarding the spacing we can't exactly set the width of the text based on the each tag text, so I discussed with Gal and set it as the min-width: 30px. so based on other tag content width its adjust. so if you want to set the different min-width for the different content also we can set it. will discuss and change it accordingly.

realdreamer:
@Maruceau 
There should be 11 pixel spacing between rows of tags, not there is no spacing. Or is this part of the query builder container and not the tag? 

-- spacing adjustments between the tags actually will do it in the query-builder-container

In the "Filter tag with delete" example, Platform is truncated and the string of platforms after "Fi...". What is the rule when to truncate? I think truncating Platform into Platfo... doesn't make a lot of sense.

There seems to be quite a bit of space between the condition and the range 1 - 5, why is that bigger than the list of platforms?

-- regarding the spacing we can't exactly set the width of the text based on the each tag text, so I discussed with Gal and set it as the min-width: 30px. so based on other tag content width its adjust. so if you want to set the different min-width for the different content also we can set it. will discuss and change it accordingly.

src/settings/_colors.scss
realdreamer: I'll check with @Maruceau because filter tag error we are using red-4 with the 20% opacity.

src/components/_filter-tags.scss
MLKiiwy: Chip -> https://material-ui.com/demos/chips/

galagmon:
@Maruceau should we maybe do the red border as hover and active of the error and not by default?

galagmon:
@Maruceau should we maybe do the red border as hover and active of the error and not by default?

galagmon:
@Maruceau should we maybe do the red border as hover and active of the error and not by default? 
I agree @realdreamer can we make sure we don't want to truncate things like platform, didn't we say that the first part is not truncated unless it reaches the max width?

galagmon:
@Maruceau should we maybe do the red border as hover and active of the error and not by default? 
I agree @realdreamer can we make sure we don't want to truncate things like platform, didn't we say that the first part is not truncated unless it reaches the max width?

realdreamer:
@galagmon it truncates only if container reaches the max-width, but at the same time the other content (right end) is more, then that takes priority. 

one solution is we can set first content area is always visible and should set max-width to the first content area. Check below screenshot for reference

screen shot 2018-12-07 at 10 50 35 am

but in this case sometimes chips container width  might not be 400px exactly. Check below screenshot for reference

screen shot 2018-12-07 at 10 50 22 am

galagmon:
@realdreamer the truncate looks good to me! Thanks 👍

Maruceau:
I updated the error state on Invision and discussed it with Gal. I think we're good to go forward with this now.

kagnamazian:
@realdreamer `There should be 11 pixel spacing between rows of tags, not there is no spacing. Or is this part of the query builder container and not the tag?` ==> yes it's cover in the query builder container. no in the tag itself

realdreamer:
@Maruceau I updated the error state. Can you please review it?

src/components/_chips.scss
```diff
@@ -0,0 +1,260 @@
+// Chips
```
kagnamazian:
```suggestion
// Chip
```
I think we can keep it singular. Since they can exist on their own

realdreamer: > When it's disabled the close arrow should not be reachable no? I can still hover it and click and my cursor change. Maybe you should remove the arrow button completely since it's disabled and no interaction can be done. 

Its a looks like a disabled state but for design wise its empty state so they can remove it instantly.

kagnamazian: ha ok. in that case we should call it empty state? let's rename it to align with design wish.

realdreamer: I don't think we should call it as a empty state (since its a visual apperanace) in stylabilla, but in react component we should maintain the state as a empty state.

galagmon:
@realdreamer is the acceptability of the chip component part of this task?

realdreamer:
@galagmon updated accessibility

PR #246, opened 2018-12-07
feat(table-list): add new variations

MLKiiwy:

Preview: https://static-staging.usabilla.com/stylabilla/pr/blup-149-table-list/docs/section-components.html#kssref-components-table-list

From design: https://projects.invisionapp.com/d/main/#/console/11358367/334094110/inspect


MLKiiwy:
@Maruceau actually the sb-text--s class force to a line-height of 19px, we have to consider creating a new class for forcing to 16px?

Maruceau:
No, don't force to 16px, leave it for now. I would like to revisit the line-height in general but we can do that at a later stage.

kagnamazian:
@MLKiiwy @Maruceau Agree let's not force that. We should refactor the component since you've been asking this line-height multiple time. We just have to be aware that it will affect existing component as well. so we must be careful.

src/components/_table-list.scss
```diff
@@ -3,13 +3,15 @@
...
-//  <ul class="sb-table-list">
+//  <ul class="sb-table-list {{modifier_class}}">
 //    <li class="sb-table-list__item"> Element 1 </li>
 //    <li class="sb-table-list__item"> Element 2 </li>
-//    <li class="sb-table-list__item"> Element 3 </li>
+//    <li class="sb-table-list__item sb-table-list__item--selected"> Element 3 </li>
 //    <li class="sb-table-list__item"> Element 4 </li>
 //  </ul>
 //
+// .sb-table-list--without-line-divider - Remove line divider between elements
```
kagnamazian: why did we add this feature? I thought that was the signature of our table list to have this bottom border.

src/components/_table-list.scss
```diff
@@ -62,3 +74,148 @@
 //
 // Styleguide Components.table-list.hover
 
+// Popover - Single line
```
kagnamazian: I don't see why in stylabilla we need to add a hybrid list element mix between popover and list. It should be just a list smaller.

src/components/_table-list.scss
```diff
@@ -62,3 +74,148 @@
...
+//        <span class="sb-text--s sb-text--milo">This is a single line item</span>
+//    </li>
+//    <li class="sb-table-list__item sb-padding-top-s sb-padding-bottom-s sb-padding-left sb-padding-right">
+//        <span class="sb-text--s sb-text--milo">This is a single line item</span>
+//    </li>
+//  </ul>
+//
+// Styleguide Components.table-list.popover.single-line
+
+// Popover - Single line with checkbox
```
kagnamazian: I also confused by this component. In anyway, we should not set the word "Popover" because I can guarantee it will be used somewhere in the future. 
It's kind of table with one column. I mean the style is exactly the table no? Should we not make it belong to the table list. I'm not sure I'm open to discussion. What do you think?

src/components/_table-list.scss
```diff
@@ -62,3 +74,148 @@
...
+//
+// Browser support: IE10+, Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+//  <ul class="sb-table-list sb-table-list--without-line-divider sb-table-list--with-checkbox">
```
kagnamazian: ha yeah ok you do use `sb-table` so my feeling it's a specific table without header and column but still belong in the table group

src/components/_table-list.scss
```diff
@@ -62,3 +74,148 @@
...
+    margin-bottom: -1px;
+  }
+}
+
+.#{$ns}table-list__label {
+  flex: 1 1 auto;
+  height: sb-px2rems($checkbox-box-size);
+}
+
+// Popover - Multiple lines
```
kagnamazian: Ideally, this behaviour should be covered without adding a section. Since we use `<li>` it can accept any element. what do you think?

src/components/_table-list.scss
MLKiiwy: Ok bad naming, here it was a simple example of a typical table-list inside a popover. But as explained in the description, I'm going to remove this example and the one with multiple lines, because there is just composition of options of stylabilla, there is no special classes associated to both examples.

kagnamazian: ok perfect. So sorry I missed this description part. But I feel if you needed a design validation you could always do it in react-stylabilla as well.
Anyway we should remove it before merging indeed :)

src/components/_table-list.scss
kagnamazian: Forget this comment so.

src/components/_table-list.scss
realdreamer:
```suggestion
// .sb-table-list--borderless - Remove line divider between elements
```
instead of without-line-border, we can prefer this.

src/components/_table-list.scss
```diff
@@ -26,39 +28,112 @@
...
 // Browser support: IE10+, Safari6+, Firefox30+, Chrome30+
 //
 // Markup:
-//  <ul class="sb-table-list">
-//    <li class="sb-table-list__item sb-flex sb-align-items-center sb-hide-child">
-//        <div class="sb-padding">Element 1</div>
-//        <svg class="sb-icon sb-child" viewBox="0 0 13 16">
-//          <path d="M7.21582935 3.881V0L13 6l-5.78417065 6V8.0345C.4864288 7.8705.77515533 12.7805 2.48823387 16-1.73999488 11.259-.8424844 3.663 7.21582935 3.881"/>
-//        </svg>
+//  <ul class="sb-table-list sb-table-list--without-line-divider sb-table-list--with-checkbox">
```
realdreamer:
```suggestion
//  <ul class="sb-table-list sb-table-list--without-line-divider sb-table-list--checkbox">
```

src/components/_table-list.scss
```diff
@@ -26,39 +28,112 @@
...
+//        <span class="sb-table-list__item-info sb-margin-left-s sb-text--s sb-text--milo sb-color-light-1">More info</span>
+//      </div>
 //    </li>
-//    <li class="sb-table-list__item sb-flex sb-align-items-center sb-hide-child">
-//        <div class="sb-padding">Element 2</div>
-//        <svg class="sb-icon sb-child" viewBox="0 0 13 16">
-//          <path d="M7.21582935 3.881V0L13 6l-5.78417065 6V8.0345C.4864288 7.8705.77515533 12.7805 2.48823387 16-1.73999488 11.259-.8424844 3.663 7.21582935 3.881"/>
-//        </svg>
+//    <li class="sb-table-list__item sb-flex sb-flex-row">
+//      <div class="sb-checkbox-without-label sb-flex sb-margin-left sb-margin-top-s sb-margin-right-s sb-margin-bottom-s">
```
realdreamer:
```suggestion
//      <div class="sb-checkbox--no-label sb-flex sb-margin-left sb-margin-top-s sb-margin-right-s sb-margin-bottom-s">
```

MLKiiwy: can't change this one, its not defined by my PR

MLKiiwy:
@realdreamer @kagnamazian @Maruceau PR updated. (naming fix, useless examples removed)

For the 16px line-height, it's going to be addressed later.

src/components/_table-list.scss
kagnamazian: ?

yarn.lock
```diff
@@ -2441,7 +2441,7 @@ debug@^4.0.0:
   dependencies:
```
kagnamazian: Did you also update the package.json and you forget to add it or it's just yarn.lock which adds other information for each package?

src/components/_table-list.scss
MLKiiwy: Oh its because design wants this variation, the list has no border

kagnamazian: are you using it already in one of our components? I really thought it was our signature to have a list with a background. if we remove that I don't really see the use case. @Maruceau @galagmon do you have another reason maybe?

MLKiiwy: I'm using it in all the list displayed in the popover.

kagnamazian: ok

galagmon: @kagnamazian I think that @Maruceau wanted to have the list items within the popover without the border to avoid clutter in these small components

PR #247, opened 2018-12-10
feat(popover): add divider sub element, fix width for date-picker

MLKiiwy:

Preview: https://static-staging.usabilla.com/stylabilla/pr/popover-adds/docs/section-components.html

- [x] set a fixed width for the popover associated to the date-picker component


src/components/_date-picker.scss
```diff
@@ -742,6 +742,18 @@
...
 }
 
+/**
+ * Date picker popover
+ */
+
+$date-picker-popover-width: 242px;
+
+.#{$ns}date-picker__popover {
+  width: sb-px2rems($date-picker-popover-width);
```
realdreamer: if we set width then it will be always fixed to the size, then why we need min-width and max-width?

src/components/_popover.scss
```diff
@@ -54,3 +54,34 @@
...
+// Browser support: IE9+ (no transitions), Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+// <button class="sb-btn sb-btn--primary">Popover anchor (trigger)</button>
+// <div class="sb-modal sb-popover--open" role="dialog" aria-labelledby="popover-content" data-kss-animation-switch="sb-popover--open,sb-popover--close" data-kss-animation-name="Popover open/close">
+//   <div class="sb-popover__backdrop" aria-hidden="true"></div>
+//   <div class="sb-popover__dialog" role="document">
+//     <div class="sb-popover__content sb-margin-top-xs" role="presentation" tabIndex="-1">
+//       <h4 id="popover-title" class="sb-padding-left sb-padding-right sb-padding-top-s sb-padding-bottom-s">Popover title</h4>
+//       <hr class="sb-popover__divider"/>
```
realdreamer: I think we don't need separate `<hr class="sb-popover__divider"/>` instead of you can use `sb-border-bottom-light-3` for h4 and p tag.

MLKiiwy:
@galagmon yes

MLKiiwy:
@realdreamer removed the unnecessary divider + just use width prop

PR #248, opened 2018-12-12
fix(chip): accessibility focus issue on chip

realdreamer:

Preview: https://static-staging.usabilla.com/stylabilla/pr/fix-chip-accessibility/docs/section-components.html


realdreamer:
@MLKiiwy as you mentioned removed browsers default outline for focus. 

@kagnamazian I'll check with @galagmon for button focus and will update based on that

kagnamazian:
@realdreamer I think we should keep as much as possible the default outline for focus. But let's see with the design team. She is aware of accessibility. I trust her on every choice she decides.

realdreamer:
@kagnamazian checked with @galagmon she's fine with button focus less visibility and removed default browser outline and changed error-state focus border color

PR #252, opened 2018-12-19
feat(popover-transitions): add transitions to popover

MLKiiwy:

Preview: https://static-staging.usabilla.com/stylabilla/pr/popover-transitions/docs/section-components.html#kssref-components-popover-content-transitions

## Done

- [x] adding all animations for the popover component
- [x] add doc + sample on utils (+ some utils)
- [x] special animation between popover suggestion and calendar view for date-picker

## Why not using max-height property for animate the popover.

As you can see here: https://codepen.io/MLKiiwy/pen/ZVpPNL
There is two way to do the collapse animation, one uses the GPU processing (faster but you have some text deformation) the other keep the text.

The max-height can work in our case, but only if you open the popover at the bottom of the anchor element, and only if the popover is big enough. 

On the other hand, using the scale property all the cases are working.


src/components/_popover.scss
```diff
@@ -49,8 +42,84 @@
...
   background-color: $sb-color-white;
   border: 1px solid $sb-color-light-3;
   border-radius: sb-px2rems($sb-radius);
   box-shadow: $sb-box-shadow-light;
+  overflow: hidden;
+  transform-origin: top;
+}
+
+.#{$ns}popover--open .#{$ns}popover__content {
+  max-height: 400px;
```
kagnamazian: you already defined this property line 45

src/components/_popover.scss
```diff
@@ -17,6 +17,9 @@
 //
 // Styleguide Components.popover
 
+$sb-easeout-popover: cubic-bezier(0, 0, 0.4, 1);
```
kagnamazian: is it not something we want to standardize to all animation. And therefore being defined in the animation part?

galagmon:
@MLKiiwy both are too fast in my opinion for the amount of animations in there

MLKiiwy:
@Maruceau @galagmon I change a bit the animation again because it was not working in some cases when I really use it on the application.

MLKiiwy:
Just waiting approval from @Maruceau

MLKiiwy:
Add the special animation between suggestion popover -> calendar view for date-picker @Maruceau

PR #253, opened 2019-01-09
fix(popover): change naming for content-animated

MLKiiwy:

Simple change in naming, see thread: https://github.com/usabilla/react-stylabilla/pull/178


MLKiiwy:
Its just the content, but it works only inside a popover.

kagnamazian:
So it's the content you should leave the content name 😀

MLKiiwy:
@kagnamazian I change to sb-popover__animate-content

PR #256, opened 2019-01-11
fix(slider): improve slider transition and number alignment

realdreamer:

Preview: https://static-staging.usabilla.com/stylabilla/pr/blup-64-fix-slider/docs/section-components.html

Note: The transition effect designers can check it from only react-stylabilla. 
So, didn't add them to review


src/components/_slider.scss
```diff
@@ -111,14 +111,15 @@ $sb-slider-label-disabled-color: $sb-color-light-3;
 }
 
 .sb-input-range__slider-container {
-  transition: left 0.3s ease-out;
+  transition: left 0.2s ease-out;
```
MLKiiwy: please use everywhere the variables: https://github.com/usabilla/stylabilla/blob/master/src/settings/_animation.scss#L2

PR #260, opened 2019-01-17
fix(table-list): if long label go multiline

MLKiiwy:

Preview: https://static-staging.usabilla.com/stylabilla/pr/fix-table-list-cb-multiline/docs/section-components.html#kssref-components-table-list-single-line-checkbox

Add some min-width/height to avoid checkbox to change of size + fix label to go multiline.


galagmon:
Great! based on what is the logic of breaking the main text and the more info?

MLKiiwy:
Difficult to say but the breaking for the "more info" is calculated automatically by the browser.
I can define a min-width / max-width for this text if you want of leave the browser decide how to display it like here.

MLKiiwy:
@Maruceau @galagmon updated I use Milo for label in the list now.

MLKiiwy:
@Maruceau If I put the number of responses on one line, it will be like a table column so when there is not a lot of responses the column will be always same size. If its ok for you I can do it.

Uhm I can change the line-height, but here we are using a label and now that mean I'm overriding all the properties of default stylabilla label in that case. I can do it but we should think about refactoring the label if we override everytime

PR #262, opened 2019-01-18
feat(button): add subtle link style button

realdreamer:

Preview: https://static-staging.usabilla.com/stylabilla/pr/blup-64-button-subtle-link/docs/section-components.html#kssref-components-button-link

Implemented new subtle link button


src/components/_button.scss
```diff
@@ -345,3 +347,64 @@ $sb-btn-padding-l: sb-rem-spacing('s') sb-rem-spacing('l');
...
 }
+
+// Link
+//
+// Button Link.
+//
+// Browser support: IE9+, Safari6+, Firefox30+, Chrome30+
+//
+// Markup:
+// <button class="sb-btn {{modifier_class}}">Button</button>
```
MLKiiwy: Can you add here: 

// <button class="sb-btn {{modifier_class}}" disabled>Button</button>

realdreamer: Initially I've also added disabled and other states as well. But, Marco mentioned that for these button links, we don't need `disabled` state (you can find it in the above comments). so i removed it.

PR #264, opened 2019-01-23
feat(inline-editable-text): add editable text input styles

ranst91-u:
Reviewers: @MLKiiwy @kagnamazian @realdreamer @ranst91-u @galagmon @Maruceau @gsbkr @michelvdhoek

Preview: https://static-staging.usabilla.com/stylabilla/pr/blup-187-inline-editable-text/docs/section-components.html#kssref-components-editable-text

Invision ref : https://projects.invisionapp.com/d/main#/console/11358367/342290540/preview


ranst91-u:
I am wondering if I should just rename this to "editable text". I initially used the name suggested on the task. However, I think the utilisation of this might suggest that the "inline" is redundant

src/components/_inline-editable-text.scss
```diff
@@ -0,0 +1,37 @@
+// Inline Editable Text
+//
+// Browser support: Safari11.1+, Firefox53+, Chrome57+
+//
+// Unsupported browsers will show default caret color
+//
+// Markup:
+//  <input type="text" class="sb-inline-editable-text" placeholder="Add a hint text to guide the user" /> <br>
+//  <input type="text" class="sb-inline-editable-text" value="This is an editable label" />
```
MLKiiwy: regarding the future usage and the implementation suggested by @Maruceau for the 1st screen of chart's creation, I would like to add:
- one example multiline
- one example with another size of text (using existing text modifier of usabilla)

realdreamer:
Great work 👍 
But, just a thought since its a input style, we can re-use the input component and modify it with classnames instead of separate `inline-editable` component. 

@MLKiiwy @kagnamazian what do you people think?

galagmon:
Can we add the Jira ticket to the first comment as well with the invision link so I can see what is the design we are referring to?

src/components/_inline-editable-text.scss
kagnamazian: Yes agree with Mikael : 
here the example of text modifier https://static-staging.usabilla.com/stylabilla/pr/blup-187-inline-editable-text/docs/section-style.html#kssref-style-typography-text

MLKiiwy:
@Maruceau @ranst91-u 
If I check invision, it seems the "empty selected" state is an impossible state, because user can't select the "placeholder" content.

Also, both focus state can change a bit when implementing in react-stylabilla:
- if we want to put the cursor at the end on focus with keyboard (and mouse?) in a filled one. That behavior is only possible with some JS.
- same if we want cursor at the begining

src/components/_editable-text.scss
```diff
@@ -0,0 +1,58 @@
...
+//  </div>
+//
+// Styleguide Components.editable-text
+
+$sb-editable-text-selection-color: $sb-color-blue-4;
+$sb-editable-text-hint-color: $sb-color-light-1;
+
+.#{$ns}editable-text {
+  display: inline-block;
+  border: none;
```
MLKiiwy: this property is not required anymore.

src/components/_editable-text.scss
```diff
@@ -0,0 +1,58 @@
...
+  outline: none;
+  caret-color: $sb-editable-text-hint-color;
+  text-align: center;
```
MLKiiwy: I feel we should add **sb-text--center** manually on the span if we want center alignement

src/components/_editable-text.scss
```diff
@@ -0,0 +1,58 @@
...
+//
+// Unsupported browsers will show default caret color
+//
+// Markup:
+//  <div class="sb-flex-row sb-justify-content-start">
+//    <div>
+//      <span contenteditable class="sb-editable-text" placeholder="Add a hint text to guide the user"></span>
+//    </div>
+//    <div>
+//      <span contenteditable class="sb-editable-text sb-title--l">This is an editable label</span>
```
MLKiiwy: can you add after the span for each example, a detail of modifier applied?
Something like:
`<div class="kss-modifier__name kss-style">.sb-title--l</div>`

+ adding a bit of space between example?

ranst91-u: There are no modifiers for this class. We have pure css reading empty or filled and changing the styles accordingly

src/components/_editable-text.scss
ranst91-u: @Maruceau What do you think? should this be a part of the style? or should we apply it when needed?

galagmon:
Hey looks great!
Missing a few things though: (let me know if they should be on react stylabilla)
- When I click on the hint text the blinking cursor should appear on the right and the line should stay no? @Maruceau 
- When I erase the text I edited and click out of the component the hint text/place holder text should come back? 
- With keyboard navigation, the placeholder text shows the blinking cursor but in the hint text it doesn't 
- Should the place holder text color be darker than the text I fill in? @Maruceau

ranst91-u:
> Hey looks great!
> Missing a few things though: (let me know if they should be on react stylabilla)
> 
> * When I click on the hint text the blinking cursor should appear on the right and the line should stay no? @Maruceau
- The darker ones are the texts, when you click them you are starting an edit from the point you just clicked (so the behaviour is more of an "editing" than "adding" (adding would be if the pointer starts at the end)
> * When I erase the text I edited and click out of the component the hint text/place holder text should come back?
- Just checked it. That is because I did not add a placeholder for existing texts in the examples. But this functionality works. Should I add it?

I gave some inline answers

src/components/_editable-text.scss
```diff
@@ -0,0 +1,63 @@
...
```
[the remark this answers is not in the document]
MLKiiwy: The main issue is the multiline support if we go for an input.
In that case, the simple solution will be to have two different markup for multiline or not. (input / textarea) And in the react implementation we use a boolean for enabling multiline or not.

In all cases, we are missing the aria labels in the examples https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/textbox_role, for accessibility we should always provide them.

MLKiiwy:
And also please rebase on master before merging

PR #267, opened 2019-01-29
fix(inline-editable-text): Add disabled editing label

ranst91-u:
Reviewers: @MLKiiwy @kagnamazian @realdreamer @ranst91-u @galagmon @Maruceau 

This is a small fix for https://github.com/usabilla/stylabilla/pull/264

The fixes are:
- display: inline-block is now only applied for span elements
- No underline will be visible for disabled editable-text components (supplied with data-disabled)

Preview: https://static-staging.usabilla.com/stylabilla/pr/fix-blup-187-inline-editable-text/docs/section-components.html#kssref-components-editable-text

Invision ref : https://projects.invisionapp.com/d/main#/console/11358367/342290540/preview


galagmon:
Great @ranst91-u thanks!
For the not editable text If the text is not editable then we shouldn't be able to also highlight it as it doesn't do anything.
For the first inline component when I click on it the blinking cursor is missing

Thanks 😬

PR #270, opened 2019-02-06
feat(disabled-table-row): add class to disable a row

chiel:

Preview: https://static-staging.usabilla.com/stylabilla/pr/blup-189-disabled-table-row/docs/section-components.html

Added a class name to disable a table row. This PR does not disable mouse interaction since the row will still need to be able to be hovered, for example. The checkbox, however, is hidden and is non-interactive.

As part of this PR I also moved the `sb-table__tr-inactive` selector to its own section since that class was being applied on the entire table while it should only be applied to a table row.


galagmon:
@chiel the link is not working

chiel:
@galagmon ah my bad, I didn't update it (thought it would replace that value automatically) - however, when I try with the branch name there it's also not working, perhaps @MLKiiwy or @realdreamer has some insight on what the URL should be?

realdreamer:
@chiel updated the url.

chiel:
Thanks @realdreamer. @galagmon, you should be able to check it out now. :)

src/components/_table.scss
```diff
@@ -282,6 +268,135 @@
...
+//                  </span>
+//               </span>
+//            </th>
+//         </tr>
+//      </thead>
+//      <tbody class="sb-text--s">
+//         <tr class="sb-table__tr {{modifier_class}}">
+//            <td class="sb-table__td sb-table__checkbox sb-text--center">
+//              <div class="sb-checkbox-without-label">
+//                <input type="checkbox" id="checkbox-table-1-{{modifier_class}}" class="sb-checkbox-without-label__input">
```
MLKiiwy:
```suggestion
//                <input type="checkbox" id="checkbox-table-row-1-{{modifier_class}}" class="sb-checkbox-without-label__input">
```

chiel: Good catch!

PR #272, opened 2019-02-07
feat(stepped-nav): add stepped nav component

chiel:
JIRA Ticket: https://usabilla.atlassian.net/browse/BLUP-198

Preview: https://static-staging.usabilla.com/stylabilla/pr/blup-198-stepped-nav/docs/section-components.html

For the sake of accessibility, we thought it best to use anchors, even if they are not yet active, since the navigation is supposed to linear, that's why the `href` is missing from links that are yet accessible.


chiel:
Discussed stuff with the designers, going to be implementing this as regular tabs in stylabilla and add more complex behavioural stuff in react-stylabilla/implementation, so I will close this for now.

PR #273, opened 2019-02-08
feat(thumbnail-selector): add thumbnail selection component

realdreamer:
Invision: https://projects.invisionapp.com/share/9NEFGAXM4#/screens/309083149_card

Preview: https://static-staging.usabilla.com/stylabilla/pr/blup-186-thumbnail-selection/docs/section-components.html#kssref-components-card

## Notes for testing:
Re-used the `card` component but added the modifier to make it fixed width `thumbnail` style. 

@Maruceau I manage it to use your breakpoints list only, you can check it now.


src/components/_card.scss
```diff
@@ -78,3 +78,255 @@ a.#{$ns}card:focus,
...
+    }
+  }
+}
+
+// Card - compact with radio button selection
+//
+// Browser support: IE11+, Safari6.1+, Firefox30+, Chrome30+
+//
+// Markup:
+// <div class="sb-thumbnail-container">
```
MLKiiwy: I think we are missing some aria labels and role in this example.
```suggestion
// <div class="sb-thumbnail-container" role="radiogroup">
```

chiel: Might even want to change it to a `fieldset` in this particular case actually (which would mean we'd need to reset the browser-default margins/paddings as well).

realdreamer: since we are not actually creating a `form` element, so I think `fieldset` wont be necessary. what do you think?

realdreamer:
@chiel the tab is working, my bad missed the focus state styling :(

src/components/_card.scss
chiel: Hmm, dunno - it's a radio group. Not 100% sure on on the semantics of `fieldset`, if it should only be inside a `form` or not.

chiel:
Ah fair enough :)

realdreamer:
> I love the way it scales!
> 
> * In the example ".sb-card--compact" I miss the line divider between the illustration and the label box. It is visible in the default styling.
`default` styling is the `.sb-card` styles since we are using most of the styles from `card` component
> * I can't seem to tab through the different thumbnails
you can use keyboard navigation only in `card input selection` tab key to focus the radio button group and use arrow keys to navigate between different thumbnails, since its a radio button
> * Like the bigger shadow on hover, however, it is inconsistent with a card. I think this should be consistent
updated 
> * there should be 16-pixel margin around the label. In the default styling, it is 23, in ".sb-card--compact" it is correct.
the `default` style inherits from general `card` component, thats why it is `23px`. `compact-card` is the one for thumbnail style.
> 
> I'm already really happy with it.

galagmon:
@realdreamer is this also fixing the issue in the overview page where the [ x responses ] is a link? we need to make sure it doesn't happen here :) 
Are the illustrations get cut on top based on the titles?
Other than that beautiful!

realdreamer:
@galagmon for the illustration getting cut on top, @Maruceau said that he'll work on it, but not now maybe after sometime..

src/components/_card.scss
```diff
@@ -72,9 +76,263 @@ a.#{$ns}card:focus,
...
+  font-family: $sb-primary-font-family;
+  font-size: sb-px2rems(13px);
+  line-height: sb-px2rems(16px);
+}
+
+@each $sb-thumbnail-breakpoint, $breakpoint in $sb-thumbnail-breakpoint-map {
+  $thumbnail-padding-size: 2 * map-get($sb-spacing-map, $sb-thumbnail-container-padding);
+  $thumbnail-margin:  2 * map-get($sb-spacing-map, $sb-thumbnail-margin);
+  $thumbnail-count: map-get($sb-thumbnail-count-map, $sb-thumbnail-breakpoint);
+  $breakpoint: map-get($sb-thumbnail-breakpoint-map, $sb-thumbnail-breakpoint);
```
chiel: This line is not needed anymore since you already get `$breakpoint` from the `@each` now.
```suggestion
```

PR #274, opened 2019-02-08
feat(tabs): add tabs component

chiel:

Preview: https://static-staging.usabilla.com/stylabilla/pr/blup-198-tabs/docs/section-components.html#kssref-components-tabs

Marco is still finalising some states, so there might be a small change in how the inactive but selectable state looks, as well as perhaps some focus styles.


chiel:
@usabilla/fe-chapter + @galagmon: currently I have a separate class for marking a tab as disabled, but I also remove the `href` attribute in order for it not to be clickable which kinda feels like overkill. I feel like one of the following would be better:

- Just remove the `href`, and style it based on the missing `href`. Potential downside of this could be that it introduces accessibility issues, I'm not sure if there are any, but it's a possibility.
- Keep the class, and use `pointer-events: none` to make it inactive. Potential downside here is kinda two-fold: since we'd be using `pointer-events: none` the tab wouldn't be interactive at all, which means we would not be able to change the cursor, or add some kind of tooltip to explain WHY the tab is disabled.

Of course we could keep doing it as I have now, but it feels like more effort than it should be. In an ideal world anchors would also support a `disabled` attribute like buttons, but alas.

Would love to hear your thoughts on this.

realdreamer:
@chiel In the first approach without `href` it will not get focus with the keyboard the same behavior use to happen even when the button is `disabled`. So, I think first approach will be fine.

chiel:
@realdreamer yeah good point. However, if the button needs to be focusable (and I think that it should), we will need to block the navigation with javascript then.

src/components/_tabs.scss
```diff
@@ -0,0 +1,79 @@
...
+// JavaScript is required to actually add/remove the href and classes and position/animate the
+// line on the active item. In order to animate the line, scale + translate should be used to
+// leverage hardware acceleration.
+//
+// Browser support: IE11+, Safari6.1+, Firefox22+, Chrome21+
+//
+// Markup:
+// <nav class="sb-tabs sb-text--smallcaps">
+//   <div class="sb-tabs__inner">
+//     <a href="#chart-type" class="sb-tabs__tab--active">Chart type</a>
```
realdreamer:
```suggestion
//     <a href="#chart-type" class="sb-tab-link sb-tab-link--active">Chart type</a>
```
giving the explicit classnames avoid the nesting in styling.

chiel: I thought about doing this but I thought it felt a bit excessive to have classes assigned even for the default links. That said, it if more explicit so probably worth it - it seems this is the way it is done everywhere anyway. :)

src/components/_tabs.scss
```diff
@@ -0,0 +1,79 @@
...
+
+  &__underline {
+    position: absolute;
+    left: 0;
+    bottom: 0;
+    width: 1px;
+    height: 2px;
+    background: $sb-color-blue-2;
+    transform-origin: 0;
+    transition: transform $sb-timing-s $sb-easeout-popover;
```
chiel: I am abusing a different easing function here cause the default one has the "elastic" effect which does not really look that nice here. Is this fine, or should we perhaps call this variable something more generic like `$sb-easeout-function`?

src/components/_tabs.scss
chiel: I've updated this now btw - I didn't accept the suggestion cause there were also style changes involved.

src/components/_tabs.scss
```diff
@@ -0,0 +1,85 @@
...
+
+.#{$ns}tabs__inner {
+  display: flex;
+  justify-content: center;
+  position: relative;
+}
+
+// specifying the tag here as well due to specificity of
+// the default link colour otherwise prevailing
+a.#{$ns}tabs__tab {
```
realdreamer:
```suggestion
.#{$ns}tabs__tab {
```

src/components/_tabs.scss
```diff
@@ -0,0 +1,85 @@
...
+  margin-left: 40px;
```
realdreamer:
```suggestion
  margin-left: sb-rem-spacing('l');
```

src/components/_tabs.scss
kagnamazian: it depends where we want to go. We should ask the Design if this effect is supposed to be the easeout transition for all our component? if yes, then you can change it totally agree.

src/components/_tabs.scss
chiel: If I remove this, all the tabs will be green (as explained per the comment above), because `a:hover, a:visited` has a higher specificity than `.sb-tabs__tab`.

src/components/_tabs.scss
chiel: Well, I talked with @Maruceau, and this is at least the easing function that we also want to use for the tabs. In the end it actually became the `easeboth` one, though.

PR #275, opened 2019-02-15
fix(card): remove shadow and movement from the active state

realdreamer:
Reviewers: @chiel @MLKiiwy @kagnamazian @realdreamer @ranst91-u @Maruceau 

Preview: https://static-staging.usabilla.com/stylabilla/pr/fix-card-active-state/docs/section-components.html

There is a small change in the component when the input is checked, active state should be similar to the default card state except `border` and `color` change of the title


realdreamer:
Input section: https://static-staging.usabilla.com/stylabilla/pr/fix-card-active-state/docs/section-components.html#kssref-components-card-radio

PR #276, opened 2019-02-15
fix(card): ie11 thumbnail image not visible

realdreamer:

Preview: https://static-staging.usabilla.com/stylabilla/pr/fix-ie11-card/docs/section-components.html

*IE11 Bug:*
The image is not visible on IE11, because the child element not inheriting the height from parent element. 

### Screenshot:

Before: 
screen shot 2019-02-15 at 4 59 49 pm

After:
screen shot 2019-02-15 at 5 00 24 pm


realdreamer:
@kagnamazian yeah, thats what I was mentioned in the morning stand-up `card` component hover itself not working in the IE11 and edge, spent some time to debug on friday, but it looks like a mystery, will have to look into it.

PR #278, opened 2019-02-19
Fix tab styling

chiel:
Preview: https://static-staging.usabilla.com/stylabilla/pr/fix-tabs/docs/section-components.html#kssref-components-tabs

Some of the colours being used were not according to design - this has been rectified. Additionally, updated the example to use an absolutely positioned underline to get some notion of what it will look like. Position might not be 100% correct on all browsers, since not all browsers render fonts exactly the same. When measured with JavaScript in `react-stylabilla`, this position + width should always be correct.

@Maruceau As @galagmon commented on the PR in `react-stylabilla`, there is currently no state for active + focussed which will need to be decided upon how this looks. This is also the case for disabled + focussed.

So, still a couple open ends, will finish this up tomorrow.

## Reviewers:
- Frontend: @usabilla/fe-chapter
- UX/UI: @usabilla/design

## Checklist minimal
- [x] Components are navigable by keyboard only (design required as described above)
- [x] Appropriate alt and [ARIA attributes](https://w3c.github.io/using-aria/) are included
- [x] Semantic HTML is used
- [ ] ~The component works across all device widths, and is built mobile first.~ (It depends on the width of the elements in the component, there is no design for how this should function on mobile/tablet)

- [x] Chrome last version
- [x] Firefox last version
- [x] Edge last version (underline alignment is a little bit off due to font rendering)
- [x] IE11 (underline alignment is a little bit off due to font rendering)
- [x] Safari 10+ (keyboard navigation does not seem to work properly, which seems to be a macOS limitation)
- [ ] ~IE9 - IE10 - Only if your element can be part of a public feedback form~ (won't be used in a public feedback form)


src/components/_tabs.scss
```diff
@@ -11,10 +11,10 @@
 // Markup:
 // <nav class="sb-tabs sb-text--smallcaps">
 //   <div class="sb-tabs__inner">
-//     <a href="#chart-type" class="sb-tabs__tab sb-tabs__tab--active">Chart type</a>
-//     <a href="#data" class="sb-tabs__tab">Data</a>
+//     <a href="#chart-type" class="sb-tabs__tab">Chart type</a>
+//     <a href="#data" class="sb-tabs__tab sb-tabs__tab--active">Data</a>
 //     <a href="#visualization" class="sb-tabs__tab sb-tabs__tab--disabled">Visualization</a>
-//     <span class="sb-tabs__underline"></span>
+//     <span class="sb-tabs__underline" style="transform: translateX(136px) scaleX(53);"></span>
```
ranst91-u: Any reason for inline styles?

chiel: Yeah, they are really only meant for the documentation, since the widths of the tabs are dependent on the text inside of them, so I don't want to put them in the styles that actually get compiled.

chiel:
@Maruceau this one is ready to be reviewed now.

PR #280, opened 2019-02-22
fix(tabs): update tabs styling

chiel:
Preview: https://static-staging.usabilla.com/stylabilla/pr/update-tabs-styling/docs/section-components.html#kssref-components-tabs

While reviewing the `react-stylabilla` implementation, @Maruceau became less and less happy with his solution for keyboard focus so we decided to update it again. After trying a couple of different things, we came to the conclusion that the underline animation introduced too many "weird" looks while transitioning between states.

As a result, the underline has been removed entirely and the border is now a pseudo element, for reasons explained in the code.

## Reviewers:
- Frontend: @usabilla/fe-chapter
- UX/UI: @Maruceau 

## Checklist minimal
- [x] Appropriate alt and [ARIA attributes](https://w3c.github.io/using-aria/) are included
- [x] Semantic HTML is used

- [x] Chrome last version
- [x] Firefox last version
- [x] Edge last version
- [x] IE11
- [x] Safari 10+
- [ ] ~IE9 - IE10~ - Only if your element can be part of a public feedback form


src/components/_tabs.scss
```diff
@@ -2,95 +2,82 @@
...
-    background: $sb-color-light-4;
+  // we don't want the underline borders to "taper off" towards the sides, which would happen because
+  // there are borders on all sides of the tab, so we use a pseudo element for the bottom border instead
+  &::after {
+    content: '';
+    position: absolute;
+    right: -1px;
+    bottom: -2px;
+    left: -1px;
+    height: 2px;
```
realdreamer: it would looks nice if we add some transition for `background-color`. what do you think?

chiel: Again, that would be up to @Maruceau - I believe he prefers the instant (dis)appearance of the underline.

chiel: He wants to see if in `react-stylabilla` first, so I will go ahead and merge this one for now.

PR #282, opened 2019-03-05
fix(card): add "locked" svg image and centered image option

ranst91-u:
Preview: https://static-staging.usabilla.com/stylabilla/pr/blup-199-chart-type/docs/section-components.html

Added a "locked" svg image in the thumbnail card options. This will be used for "Coming soon" chart types

## Reviews

I assume the rest of the checks are not needed, since it's just a small fix


src/components/_card.scss
```diff
@@ -172,7 +172,7 @@ $sb-thumbnail-desc-color: $sb-color-light-1;
 $sb-thumbnail-title-color: $sb-color-dark-2;
 $sb-thumbnail-title-active-color: $sb-color-blue-2;
 
-$sb-thumbnail-image-list: mood nps opentext responses rating;
+$sb-thumbnail-image-list: mood nps opentext responses rating locked;
```
realdreamer:
```suggestion
$sb-thumbnail-image-list: mood nps opentext responses rating lock;
```
or `coming-soon` something like

chiel: When it comes to iconography, I would just pick a name that represents what the icon is - the context in which it is used should only matter to naming if it is named within that context.

ranst91-u: so "lock" or "coming soon"?

MLKiiwy: lock 👍

PR #291, opened 2019-03-26
feat(header): refactor header component

ranst91-u:
Preview: https://static-staging.usabilla.com/stylabilla/pr/blup-364-header-component/docs/section-components.html#kssref-components-header

Redesign of header component, and change of back button.
Mobile and device width support notice:
Mobile widths are basically supported, however, as the component is filled with items and button, there might be mobile view issues

## Reviews

## Checklist minimal
- [x] Semantic HTML is used

- [x] Chrome last version
- [x] Firefox last version
- [x] Edge last version
- [x] IE11
- [x] Safari 10+


src/components/_header.scss
```diff
@@ -6,83 +6,64 @@
 //
 // Markup:
 // <header class="sb-header">
```
kagnamazian: could we add a border maybe around the header to see the size of the header? Otherwise, it's impossible to see the difference between the header and the tile.
It's just for the documentation purpose

src/components/_header.scss
```diff
@@ -6,83 +6,64 @@
...
+  box-sizing: border-box;
+  flex-direction: row;
+  flex: 1;
 }
 
-.#{$ns}header__btn {
-  margin-top: sb-rem-spacing('s');
+.#{$ns}header__center {
+  overflow: hidden;
+  flex-direction: row;
```
kagnamazian: you do not need to repeat the flex direction on the children `flex-direction: row`. Also it's the default value of flex-direction. so you don't need to add it at all : https://www.w3schools.com/cssref/css3_pr_flex-direction.asp

src/components/_header.scss
```diff
@@ -6,83 +6,64 @@
...
 }
 
-.#{$ns}header__title,
-.#{$ns}header__info {
-  margin-right: sb-rem-spacing('s');
+.#{$ns}header__start {
+  min-width: sb-rem-spacing('xl');
+  padding: sb-rem-spacing('xxs');
+  box-sizing: border-box;
+  flex-direction: row;
```
kagnamazian: same here

src/components/_header.scss
```diff
@@ -6,83 +6,64 @@
...
-  line-height: 0;
-
-  &:focus,
-  &:hover,
-  &:active {
-    cursor: pointer;
-    background-color: $sb-color-light-4;
-    outline: none;
-  }
+  flex-direction: row;
```
kagnamazian: same here

src/components/_header.scss
```diff
@@ -5,84 +5,61 @@
...
 }
 
-.#{$ns}header__title,
-.#{$ns}header__info {
-  margin-right: sb-rem-spacing('s');
+.#{$ns}header__start {
+  min-width: sb-rem-spacing('xl');
+  padding: sb-rem-spacing('xxs');
+  box-sizing: border-box;
+  flex: 1;
```
chiel: The header should not grow or shrink at all, would be best to set this to `flex: 0 0;`, same for the end, and then have `flex: 1 0;`for the center one. That way only the center thing will grow at all.

ranst91-u: The end should grow, as it might have several items inside

chiel: Yeah, but that's not what flex-grow means - flex-grow just operates on the REMAINING space inside an element, after all elements have taken up all the space they require, so if you set it to 0 it will still grow to accommodate everything you want to place inside of it, but will not grow to take up additional space. :)

src/components/_header.scss
```diff
@@ -5,84 +5,61 @@
...
 }
 
-.#{$ns}header__btn {
-  margin-top: sb-rem-spacing('s');
+.#{$ns}header__center {
+  overflow: hidden;
+  flex: 100;
```
realdreamer:
```suggestion
  flex: 1 0 0;
```
if you are trying to grow the center element

ranst91-u: should `flex: 1` do?

chiel: Yes, `flex: 1` is enough since you really only need to specify `flex-grow`.

src/components/_header.scss
```diff
@@ -5,84 +5,60 @@
...
-.#{$ns}header__text {
-  display: flex;
-  align-items: flex-start;
-  flex-direction: column;
+.#{$ns}header__title {
+  margin-right: sb-rem-spacing('s');
+  text-overflow: ellipsis;
+  overflow: hidden;
+  white-space: nowrap;
+  flex: 1;
```
chiel: I don't think this is actually needed - if you wanted to put something right next to the title you wouldn't be able to, since the fact that it grows would push everything else next to the end section.

src/components/_header.scss
ranst91-u: So I can mark this resolved

PR #292, opened 2019-03-27
fix(header): replace minimum height with padding for ie11 support

ranst91-u:
Preview: https://static-staging.usabilla.com/stylabilla/pr/blup-364-header-fix/docs/section-components.html#kssref-components-header

Elements with minimum height introduce issues with flex alignment on ie11. The minimum height is replaced with padding to fix this issue while preserving the style of the header

## Reviews

## Checklist minimal
Checklist is not required for this small fix. Browser compatibility full


galagmon:
image
There is one pixel on the right side. do you think it will affect the alignment in analysis app?

PR #296, opened 2019-04-01
fix(modal): add extra spacing for the header

realdreamer:
Preview: https://static-staging.usabilla.com/stylabilla/pr/fix-modal-header/docs/section-components.html#kssref-components-modal

Added some extra spacing for the modal header `11px` in the top, @Maruceau specifically requested for it. 

## Reviews

Reviewers: @chiel @MLKiiwy @kagnamazian @realdreamer @ranst91-u @Maruceau @galagmon 

## Checklist minimal
N/A

N/A


src/components/_modal.scss
```diff
@@ -12,7 +12,7 @@
 //       <button class="sb-modal__close sb-btn sb-btn--subtle" aria-label="Close">
 //         <span class="sb-icon sb-icon--close" aria-hidden="true">&times;</span>
 //       </button>
-//       <h1 class="sb-title" id="modalTitle">Modal header</h1>
+//       <h1 class="sb-title sb-modal__title" id="modalTitle">Modal header</h1>
```
kagnamazian: why not adding sb-padding-top-s?

realdreamer: if its the `sb-padding-top-s` then we can add directly in the analysis-app itself and also it will become like helper class not related to the modal-title.

realdreamer: yeah, it makes sense already other class is helper only. so, its better to go with `sb-padding-top-s`

PR #298, opened 2019-04-03
fix(tooltip): set font-variant to none

realdreamer:

In our use case, tooltip is within the tab, since tab have `font-variant: small-caps` tooltip content is also becoming so. to override added this property.

## Reviews

## Checklist minimal
N/A

N/A


realdreamer:
@kagnamazian when we refactor the tooltip, then it will become the portal like chiel mentioned, then it will not inherit from any parent styles. 

feat vs fix even I don't have a strong opinion for it.

PR #303, opened 2019-04-10
feat(chips-container): add ChipsContainer component

chiel:
Preview: https://static-staging.usabilla.com/stylabilla/pr/chips-container/docs/section-components.html#kssref-components-chipscontainer

The react-stylabilla `ChipsContainer` component depends on adding classes to its children, however if the child is a fragment the classes do not get applied correctly. This approach adds margins to the direct children of the `ChipsContainer` without the need to add additional classes.

## Reviews

## Checklist minimal
- [x] Semantic HTML is used

- [x] Chrome last version
- [x] Firefox last version
- [x] Edge last version
- [x] IE11
- [x] Safari 10+
- [ ] ~IE9 - IE10 - Only if your element can be part of a public feedback form~


chiel:
@Maruceau it does belong here, I'll have a look. I'm using `s` spacing, which **should** be 11px, but for some reason it's not.

PR #306, opened 2019-04-23
fix(toast): small design tweaks

chiel:
Preview: https://static-staging.usabilla.com/stylabilla/pr/toast-fixes/docs/section-components.html#kssref-components-toast

Some small design tweaks:

- Since it's now the container responsible for positioning the toasts, it should also have the correct `z-index`
- Since we don't really support anything below 960px for now, the tweaks for the `breakpoint-l` are now included in the default styling and the breakpoint has been removed
- Toasts now have a fixed width of 550px to ensure they are visually consistent

## Reviews

## Checklist minimal
- [x] Semantic HTML is used
- [ ] ~The component works across all device widths, and is built mobile first.~

- [x] Chrome last version
- [ ] Firefox last version
- [ ] Edge last version
- [ ] IE11
- [ ] Safari 10+


src/components/_toast.scss
```diff
@@ -49,11 +48,10 @@
   align-items: stretch;
   background: $sb-color-white;
   margin-bottom: sb-rem-spacing('s');
-  min-width: sb-px2rems(250px);
+  width: sb-px2rems(550px);
```
realdreamer: is it `550px` or `500px`?

chiel: 550 - I updated the description haha.

chiel:
@ranst91-u they should ideally all use the z-indices as defined in the settings, that way we can regulate this kinda stuff well.

PR #308, opened 2019-04-25
fix(modal): consider small screens when positioning bigger modals

ranst91-u:
Preview: https://static-staging.usabilla.com/stylabilla/pr/blup-451-modal/docs/section-components.html

The change of position (translate values) is pushing big modals (almost 600px height - our minimum height screen) above the top of the view. Removing these help displaying the modal unclipped for all screen sizes, and for all modals.

To better understand the fix, here is an example of the issue:
over-top

and the solution:
centered

## Reviews


MLKiiwy:
But there is no animation anymore?

ranst91-u:
> But there is no animation anymore?

Not sure about the animations, I never touched them.

MLKiiwy:
forget it, my mistake it's good

PR #313, opened 2019-05-09
feat(interactive-btn): add interactive button style class

ranst91-u:
Preview: https://static-staging.usabilla.com/stylabilla/pr/blup-499-subtle-interactive-btn/docs/section-components.html#kssref-components-button-colors

Add "interactive button" style. Which should currently by used by several subtle buttons

## Reviews

Reviewers: @chiel @MLKiiwy @thefrontendwizard @kagnamazian @realdreamer @DavideBruner @ranst91-u @Maruceau @gsbkr 


src/components/_button.scss
```diff
@@ -216,6 +216,7 @@ $sb-btn-link-padding: sb-rem-spacing('xs');
 // .sb-btn--danger - Danger button
 // .sb-btn--warning - Warning button
 // .sb-btn--success - Success button
+// .sb-btn--interactive - Success button
```
realdreamer:
```suggestion
// .sb-btn--interactive - Interactive button
```

maybe primary outline something like that, all that buttons are interactive right!

ranst91-u: When I talked to @Maruceau that was the name it was given.
I agree we might need to discuss the naming and decide. But it's defintely not primary outline. It wlll currently by applied to some subtle buttons

ranst91-u:
@realdreamer @Maruceau 
I resolved Mohamed's suggestion, however, if the naming discussion should still take place, I suggest we have it.
Currently, the name we use for this is "interactive", to differentiate it from normal subtle buttons (this currently applied to subtle buttons). Let me know if anything needs changing or if I should give it a stronger binding to the subtle button

Maruceau:
I see this is merged already, so I don't if my opinion still matters. But why do we need this? All buttons are interactive so I don't understand why we have a separate button called interactive.

PR #315, opened 2019-05-15
fix(chart thumbnail): remove chart thumbnail images and use only placeholder images

realdreamer:
Preview: https://static-staging.usabilla.com/stylabilla/pr/blup-571-update-chart-images/docs/section-components.html#kssref-components-card-compact

~Updated the chart images for all chart types.~ removed all specific chart images related to workspace and kept only placeholder image 

I also made small changes in the example, so exactly how it looks in the only one line. since we already have an example how it looks for multi-lines.

## Reviews

- [x] Chrome last version
- [ ] Firefox last version
- [ ] Edge last version
- [ ] IE11
- [ ] Safari 10+


galagmon:
@realdreamer why are the images updated in stylabilla? isn't it a place holder there? shouldn't the actual image be in the implementation?

realdreamer:
@galagmon good question :) but previously itself we are having it in stylabilla only. so I just replaced it. 
if its a concern we shouldn't have it here. then will move it in analysis-app

realdreamer:
@galagmon maybe I can still keep these images here, but I can remove it from the examples. there will have just placeholder. so images and style still served from `stylabilla`

galagmon:
@realdreamer Thanks! I think it should not be attached to stylabilla :) so they can stay as place holders but change them in analysis app

galagmon:
> Oh actually @galagmon is asking to move them completely to analysis-app if I'm right. Make quite of sense if this images are not shared between project

So I agreed with @realdreamer to move the specific images to the analysis app and have a place holder for any image in stylabilla 
We will have many different images there in the future

realdreamer:
@galagmon @MLKiiwy updated the PR. Please review it.

PR #318, opened 2019-05-22
feat(tooltip): use open/close classes to force tooltip state

ranst91-u:
Preview: https://static-staging.usabilla.com/stylabilla/pr/blup-603-tooltip-open-close/docs/section-components.html#kssref-components-tooltip-open

The tooltip now has two more modifier classes, open and close. These classes are used to force open/close state of the tooltip and override the hove/focus behaviour.

## Reviews

Reviewers: @chiel @MLKiiwy @thefrontendwizard @kagnamazian @realdreamer @DavideBruner @ranst91-u


ranst91-u:
> Generally looks fine to me, but since these classes just modify the `opacity`, won't this mess with the mouse cursor?

it looks like the logic behind showing the tooltip is all about opacity. We may decide to go a different way with this, but that's for the major refactor

chiel:
Fo shooooo, fine by me. 🚢

PR #322, opened 2019-05-31
feat(badge): add new extra small size badge

realdreamer:
Preview: https://static-staging.usabilla.com/stylabilla/pr/blup-583-badge/docs/section-components.html#kssref-components-badge

- Add new extra small size badges
- Modified the line-height of small and extra large size badges for rounded-corner badge to keep the height as same as the width. 

## Reviews

## Checklist minimal
- [ ] Export the component by importing it in the index.scss
- [ ] Components are navigable by keyboard only
- [ ] [Accessibility (ARIA attributes)](https://w3c.github.io/using-aria/) are included
- [] Semantic HTML is used

- [x] Chrome last version
- [ ] Firefox last version
- [ ] Edge last version
- [ ] IE11
- [ ] Safari 10+


src/components/_badge.scss
```diff
@@ -30,30 +34,42 @@ $extra-large-min-height: 83px;
...
+  width: sb-px2rems($large-min-width);
   padding: sb-rem-spacing('xs');
 }
 
 .#{$ns}badge--xl {
-  @include sb-text-xl();
-
-  min-height: sb-px2rems($extra-large-min-height);
-  min-width: sb-px2rems($extra-large-min-width);
+  font-size: sb-px2rems(34px);
```
thefrontendwizard: can we start using `scale-to-px` here?

thefrontendwizard: the approximate value would be `4` in the scale.

src/components/_badge.scss
```diff
@@ -30,30 +34,42 @@ $extra-large-min-height: 83px;
...
+  line-height: 1;
```
thefrontendwizard: Shouldn't line-height be `1.2` here?

realdreamer: No, checked with Marco. its 1

thefrontendwizard: You're right. Because it's centered it's not needed indeed.

src/components/_badge.scss
realdreamer: instead of going something like `4` which we might not know, whats the actual size we are trying to pass. instead of I would go for `px-2rems` so it clearly says the size.

and also instead of manually trying to figure out whats the scale value needs to be passed. I would give it to the util to handle it out.

PR #323, opened 2019-06-07
feat: add scale-to-px and scale-to-rem functions

thefrontendwizard:
Reviewers: @chiel @MLKiiwy @thefrontendwizard @kagnamazian @realdreamer @moroale93 @DavideBruner @ranst91-u


src/tools/_functions.scss
```diff
@@ -12,6 +12,40 @@
...
+@function scale($level) {
+  @return pow($sb-scale-ratio, $level);
+}
+
+// Returns the given scale level in px units
+@function scale-to-px($level) {
+  @return scale($level) * 16px;
+}
+
+// Returns the given scale level in rem units
```
ranst91-u: In these comments, you describe the only thing that function name suggests (so I can understand this part just by reading the function name).
The main idea behind the request for comment here is that if I view this function later, I can answer the following questions: "Why would I even need to scale the level to rems, and what are these levels? what is this argument"

thefrontendwizard: would it be better if I added examples like `(e.g. font-size: scale-to-rem(2))`?

ranst91-u: Yes, that would probably do well. 
You can try something like (just typing quickly from the top of my head):
"Converts the scale from the scale list for our sizes (or whatever) in rem".

Since you introduced these functions that rely on that scale list, you can simply make a bigger comment on top of the two (the pow function isn't relevant here) describing the entire thing with the list of sizes/scale levels, all the "why"s. With that, you wouldn't need the function specific comments

thefrontendwizard: There's really no list of sizes here, it's a continuous function. We should only use integers but it actually works from -infinity to +infinity. I'll add a comment that we should use integers as levels.

thefrontendwizard: Well, due to sass loop limitations, it only works with integers anyway so there's that.

PR #325, opened 2019-06-12
refactor: new tooltip component

thefrontendwizard:
BREAKING CHANGE: The old tooltip component is deprecated and these
styles only make sense in the context of our react-stylabilla private
react-stylabilla library.

These are the styles required for the react-stylabilla tooltip.

## Reviews

Reviewers: @chiel @MLKiiwy @thefrontendwizard @kagnamazian @realdreamer @moroale93 @DavideBruner @ranst91-u

Please add the group usabilla/design, only when PR is ready to be reviewed by UX/UI

## Checklist minimal
- [x] Semantic HTML is used

- [x] Chrome last version
- [x] Firefox last version
- [x] Edge last version
- [x] IE11
- [x] Safari 10+


src/components/_tooltip.scss
```diff
@@ -1,238 +1,124 @@
...
-  margin-top: -7px;
-  transition: opacity $sb-timing-s ease-out, margin $sb-timing-m $sb-easeup-function;
-
-  & > * {
-    font-size: sb-px2rems(13px);
+  box-shadow: 0 8px 4px -6px rgba(0, 0, 0, 0.1), 0 1px 4px -2px rgba(0, 0, 0, 0.3);
+  font-family: $sb-primary-font-family;
+  font-size: scale-to-rem(-1);
+  line-height: $sb-scale-ratio;
+  z-index: 180;
```
realdreamer: we already have `z-index` defined in terms of order.  https://github.com/usabilla/stylabilla/blob/master/src/settings/_z-index.scss#L8 maybe you can define one here for tooltip as well.

src/components/_tooltip.scss
```diff
@@ -1,238 +1,124 @@
...
+  opacity: 1;
```
realdreamer: `opacity` really needed here?

src/components/_tooltip.scss
```diff
@@ -1,238 +1,124 @@
...
+  }
 
-.#{$ns}tooltip--s {
-  white-space: nowrap;
-  line-height: sb-px2rems(13px);
-  padding: sb-rem-spacing('xs');
-}
+  // sizes
+  &--small {
+    padding: scale-to-px(-4);
```
realdreamer: in general we use `rem` units instead of `px`. is there any specific reason for using `px`?

src/components/_tooltip.scss
```diff
@@ -1,238 +1,124 @@
...
-//     <div class="sb-tooltip__arrow"></div>
-//   </div>
-// </div>
-//
-// Styleguide Components.tooltip.large
+  // arrow styles
+  &__arrow {
+    /* stylelint-disable-next-line declaration-no-important */
+    font-size: scale-to-px(2) !important;
+    margin: 3px;
```
realdreamer: you can use `scale-to-px` or `scale-to-rem`

src/components/_tooltip.scss
thefrontendwizard: yes, for animation purposes.

src/components/_tooltip.scss
thefrontendwizard: I'll switch to rem.

src/components/_tooltip.scss
thefrontendwizard: I'm removing this code in order for a better positioning strategy and I'll fix the gutter as well.

src/components/_tooltip.scss
thefrontendwizard: will do.

realdreamer: still I see the hardcoded value only for the `z-index`. Did you push?

src/components/_tooltip.scss
chiel: You don't actually need to set it since default opacity is 1 anyway, no?

src/components/_tooltip.scss
```diff
@@ -1,238 +1,133 @@
...
+      position: absolute;
+    }
+
+    .fill {
+      /* stylelint-disable-next-line declaration-no-important */
+      fill: $sb-color-white !important;
+    }
+
+    .stroke {
+      /* stylelint-disable-next-line declaration-no-important */
```
realdreamer: its not needed now. there is no `important` used

src/components/_tooltip.scss
```diff
@@ -1,238 +1,133 @@
...
-  margin-top: -7px;
-  transition: opacity $sb-timing-s ease-out, margin $sb-timing-m $sb-easeup-function;
-
-  & > * {
-    font-size: sb-px2rems(13px);
+  box-shadow: 0 8px 4px -6px rgba(0, 0, 0, 0.1), 0 1px 4px -2px rgba(0, 0, 0, 0.3);
+  font-family: $sb-primary-font-family;
+  font-size: scale-to-rem(-1);
+  line-height: $sb-scale-ratio;
+  z-index: 180;
```
realdreamer: its not updated yet.

src/components/_tooltip.scss
DavideBruner: Apart from this comment by Mohamed, with whom I agree, the rest looks good to me.

PR #332, opened 2019-06-24
fix(datepicker): change input width to accommodate longer date

ranst91-u:
Preview: https://static-staging.usabilla.com/stylabilla/pr/blup-802-datepicker-input/docs/section-components.html#kssref-components-date-picker

The 80px fixed width set for inputs of the date picker does not accommodate larger date inputs like "25 May 2019". It was tweaked to include a slightly larger date

## Reviews

Reviewers: @chiel @MLKiiwy @thefrontendwizard @kagnamazian @realdreamer @moroale93 @DavideBruner @ranst91-u @galagmon @Maruceau @gsbkr @michelvdhoek 


ranst91-u:
Any sort of zoom will not cause it to break.
As for dates. I tried different extra dates. Hopefully, this will accommodate to all possible 365 days of a year with several years ahead, otherwise, we have to manually select all dates in all years from now till forever

src/components/_date-picker.scss
```diff
@@ -367,7 +367,7 @@
 }
 
 .#{$ns}date-picker__input {
-  width: sb-px2rems(80px);
+  width: sb-px2rems(84px);
```
thefrontendwizard: can this be changed to `scale-to-rem(9)`?

PR #336, opened 2019-07-08
fix(badge): change scale and font sizes for badge

ranst91-u:
Preview: https://static-staging.usabilla.com/stylabilla/pr/badge-sizes-fix/docs/section-components.html#kssref-components-badge

Change of sizes for the badges:
- XS size received a larger scale point
- M size received a smaller font
- Badge line-height should now also align with its scale size.
- Removed `tabular-nums` font-variant for badges

## Reviews

Reviewers: @chiel @MLKiiwy @thefrontendwizard @kagnamazian @realdreamer @moroale93 @DavideBruner @ranst91-u @galagmon @Maruceau @gsbkr @michelvdhoek


ranst91-u:
This specific change is about badges. Therefore I would assume (and otherwise we have a different issue) that everything under badge styles is only used on NPS and rating badges.

PR #338, opened 2019-07-30
feat(button-subtle-lifted): add subtle-lifted button variant

ranst91-u:
Preview: https://static-staging.usabilla.com/stylabilla/pr/subtle-lifted-button/docs/section-components.html#kssref-components-button

Added a "subtle--lifted" button

## Reviews

Reviewers: @usabilla/fe-chapter @usabilla/design 


src/components/_button.scss
```diff
@@ -22,6 +22,7 @@
 // .sb-btn--primary - Primary button
 // .sb-btn--secondary - Secondary button
 // .sb-btn--subtle - Subtle button
+// .sb-btn--subtle--lifted - Subtle lifted button
```
realdreamer: imo it can be `sb-btn--subtle-lifted` its not actually a modifier for `subtle`

ranst91-u: It depends on how you look at it.
We wanted to get the look of a hovered subtle button as default and add on it

PR #342, opened 2019-08-14
Blos 279 feat switch

realdreamer:
Preview: https://static-staging.usabilla.com/stylabilla/pr/blos-279-feat-switch/docs/section-components.html#kssref-components-switches

- New `switch` toggle component

## Reviews

Reviewers: @usabilla/fe-chapter @galagmon @IreneFalgueras 

## Checklist minimal
- [x] Semantic HTML is used
- [ ] Example for each modifier with description

- [x] Chrome last version
- [x] Firefox last version
- [ ] Edge last version
- [x] IE11
- [x] Safari 10+


IreneFalgueras:
There seems to be a shadow inside the toggle, could it be removed?
 Please heck the attachments for reference. 
On Stylabilla
Screenshot 2019-08-14 at 16 40 13

On InVision
Screenshot 2019-08-14 at 16 40 18

galagmon:
One question @realdreamer do we need different sizes? In the form filter, it is smaller than the implementation in stylabilla.
The one in Stylabilla still makes sense as the default 23px 
but we should have a S (16px) for this one

realdreamer:
@galagmon Updated with small size variant

realdreamer:
@galagmon - I'm merging this pr for now. Once you're back, if we need to do any fine tunings will do it as a separate fix.

PR #346, opened 2019-09-02
feat(fonts): use external fonts from s3 cdn

DavideBruner:
Preview: https://static-staging.usabilla.com/stylabilla/pr/{$BRANCH_NAME}/docs/section-components.html

Because currently the fonts are part of multiple projects, we’d not have them duplicated per project , so we thought to just upload them once to the S3 bucket, and load the font-face using the cdn path.

## Reviews

Reviewers: @usabilla/fe-chapter

Please add the group usabilla/design, only when PR is ready to be reviewed by UX/UI

## Checklist minimal
- [ ] Export the component by importing it in the index.scss
- [ ] Components are navigable by keyboard only
- [ ] [Accessibility (ARIA attributes)](https://w3c.github.io/using-aria/) are included
- [ ] Semantic HTML is used
- [ ] Example for each modifier with description

- [ ] Chrome last version
- [ ] Firefox last version
- [ ] Edge last version
- [ ] IE11
- [ ] Safari 10+


src/settings/_namespace.scss
```diff
@@ -2,3 +2,6 @@
 
 // Namespace (all classes should be prefixed with sb, which stands for Stylabilla)
 $ns: 'sb-' !default;
+
+$sb-assets-path: 'assets/' !default;
+$sb-fonts-assets-path: 'https://static.usabilla.com/stylabilla/v20.8.0/assets/fonts/' !default;
```
DavideBruner: We still have to replace the url with a def one.

DavideBruner:
Checking the styleguide backlog I found that there was already a plan to improve the import of the fonts https://usabilla.atlassian.net/browse/STYLE-54

chiel:
Yeah, should poke infra (perhaps on that thread) to see if we can get the ball rolling. I'm definitely in favour of having the fonts centralised with all projects using the exact same copy.

DavideBruner:
> Yeah, should poke infra (perhaps on that thread) to see if we can get the ball rolling. I'm definitely in favour of having the fonts centralised with all projects using the exact same copy.

I'm gonna poke them again!

DavideBruner:
Proposal: https://usabilla.atlassian.net/wiki/spaces/DEV/pages/1012793349/Stylabilla+Assets+-+CDN+Proposal

